PR #10, opened 2016-07-20
also install nginx conf file, so dev-servers can run correctly

j1z0:


hitz:
This look innocous enough, I will do a test buildout here when I get a chance and merge it.

PR #20, opened 2016-10-28
settings are named same as index, so this line will fail if snovault.elasticsearch.index is set to anything other than snovault

j1z0:


hitz:
This was merged into dev with https://github.com/ENCODE-DCC/snovault/pull/19

PR #23, opened 2017-01-23
Replace copy.deepcopy with something faster

davisagli:
With this change, indexing an ENCODED demo instance took 6 hours (compared to 7-8 which Ben said is typical)


hitz:
tracked in issue #24

PR #27, opened 2017-03-23
J1z0 magic fix

hitz:
local PR of old branch from 4DN


hitz:
@j1z0 I just notice that this line never got merged, but there is no default for access-key... Do you guys actually use this?  Can we set a default?

Also load_test_data in this branch doesn't actually accept access_key?

j1z0:
@hitz, load_test_data we changed in our front end I didn't make the change to snowflake, I can add that I guess.  We do you the access key param, basically that controls auto-generation of an access key for admin which on localhost is written to ~/.keypairs.json or alternatively encrypted and stored in s3, which is then grabbed by tibanna our pipeline runner, so it has access to update metadata in response to pipeline runs.... a default of localhost would be fine...

hitz:
This was merged.

PR #29, opened 2017-04-10
override 'x-tar' mimetype within mimetypes module

Bek:


hitz:
Rejecting, x-tar may be "non standard" but it's better than assigning gzip to non-gzipped files.

PR #37, opened 2017-06-05
Changed maxtasks to 1 and chunksize to 1024.  Also added code to ensu…

tdreszer:
For mpindexer efficiency... reduced maxtasks and increased chunksize.

### DIFF
--- a/src/snovault/elasticsearch/mpindexer.py
+++ b/src/snovault/elasticsearch/mpindexer.py
@@ -110,8 +110,8 @@ def update_object_in_snapshot(args):
 # Running in main process
 
 class MPIndexer(Indexer):
-    chunksize = 32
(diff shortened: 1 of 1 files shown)


hitz:
so if I read this correctly:
it basically takes the number of uuids in the indexer batch and divides them equally across the N processors (16 in production), unless there are fewer than 1024 per (ca. 16K).

Does that mean that it if you give it <1024 it only uses one thread?

tdreszer:
Should be spreading the uuids evenly among all processes but capping that to 1024.  If proccess are 16 and uuids are 32 then chunkiness will be int((32 - 1)/16) + 1 = 2.  That is 2 uuids each for 16 worker processes.

hitz:
oh that makes more sense I misread line 136

PR #42, opened 2017-06-29
5159 cache frequently embedded objects

Bek:


hitz:
Satisfies ENCD-3356 ENCD-3348 and ENCD-2613

PR #47, opened 2017-09-20
Various changes to handle adding/editing child objects of abstract types (ENCD-2164)

davisagli:
* Added a new `form` frame which includes the object's `@type`
  so that the frontend can determine which schema to use for form building.
* Added a `_subtypes` mapping to the `/profiles/` view so that the
  frontend can provide a menu of options for what subtype of child to add.
* When putting an object that incldues an update to an existing child object,
  look up the existing object's `@type` and use that schema for validation,
  rather than the abstract type schema.
* When putting an object that includes a new child object,
  if the linkFrom property for that child uses an abstract type,
  the child must specify a `@type` so that the backend knows what
  schema to use for validation and which collection to add it to.
* Fix `update_children` to set the link value correctly when it is an
  array of multiple links (this fixes a bug where saving a child
  would clobber its existing links and store an invalid value).
* Fix the validation message shown if the user doesn't have permission
  to add a child object.

(I should work on tests before this is ready.)

### DIFF
--- a/src/snovault/resource_views.py
+++ b/src/snovault/resource_views.py
@@ -277,3 +277,11 @@ def item_view_edit(context, request):
         )
 
     return properties
+
+
+@view_config(context=Item, permission='edit', request_method='GET',
+             name='form', decorator=etag_tid)
+def item_view_form(context, request):
+    properties = item_view_edit(context, request)
+    properties['@type'] = context.jsonld_type()
+    return properties
--- a/src/snovault/schema_utils.py
+++ b/src/snovault/schema_utils.py
@@ -135,10 +135,10 @@ def linkFrom(validator, linkFrom, instance, schema):
     collections = request.registry[COLLECTIONS]
 
     linkType, linkProp = linkFrom.split('.')
+    linkCollection = collections[linkType]
     if validator.is_type(instance, "string"):
-        base = collections[linkType]
         try:
-            item = find_resource(base, instance.replace(':', '%3A'))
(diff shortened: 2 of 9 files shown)


src/snovault/schema_utils.py
```diff
@@ -174,11 +213,12 @@ def linkFrom(validator, linkFrom, instance, schema):
         if validator._serialize:
```
hitz: I don't fully understand what this is doing, but I didn't understand before you modified it either.

src/snovault/resource_views.py
```diff
@@ -277,3 +277,11 @@ def item_view_edit(context, request):
         )
 
     return properties
+
+
+@view_config(context=Item, permission='edit', request_method='GET',
+             name='form', decorator=etag_tid)
+def item_view_form(context, request):
+    properties = item_view_edit(context, request)
+    properties['@type'] = context.jsonld_type()
```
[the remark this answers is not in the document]
davisagli: I'm not aware of any, but I wasn't sure what code out there might be relying on the edit frame to only return properties that can be PUT back. I guess it would break submitting the edit-json form.

src/snovault/schema_utils.py
davisagli: I think this is just a flag for whether the validator is supposed to store the serialized, validated value in validator._validated or not. I don't know why that's not desired in some cases, but we're just conforming to the internals of the `jsonschema` library.

PR #50, opened 2017-10-06
Encd 3494 create secondary indexer

tdreszer:
Please take secondary indexer ENCD-3494.


src/snovault/elasticsearch/indexer.py
```diff
@@ -16,26 +16,282 @@
...
+        return self.put_obj(id, { 'list': a_list, 'count': len(a_list) })
+
+    def get_diff(self,orig_id, subtract_ids):
+        result_set = set(self.get_list(orig_id))
+
+        if len(result_set) > 0:
+            for id in subtract_ids:
+                subtract_list = self.get_list(id)
+                if len(subtract_list):
+                    result_set = orig_set.difference_update(set(subtract_list))
```
hitz: This line has undefined value, orig_set

src/snovault/elasticsearch/indexer.py
```diff
@@ -16,26 +16,282 @@
...
 
+class IndexerState(object):
+    # Keeps track of uuids and indexer state by cycle.  Also handles handoff of uuids to followup indexer
+    def __init__(self, es, key, title='primary'):
+        self.es = es
+        self.key = key  # "indexerstate"
+        self.title           = title
+        self.state_id        = self.title + '_indexer'       # State of the current or last cycle
+        self.todo_set        = self.title + '_in_progress'   # one cycle of uuids, sent to the Secondary Indexer
+        #self.failed_set      = self.title + '_failed'
```
hitz: I guess these were for the defunct redis version?  Probably should just drop them.

src/snovault/elasticsearch/indexer.py
```diff
@@ -16,26 +16,282 @@
...
+    def add_errors(self, errors, finished=True):
+        '''To avoid 16 worker concurency issues, errors are recorded at the end of a cycle.'''
+        uuids = [err['uuid'] for err in errors]  # better be uuids!
+        if len(uuids) > 0:
+            # Forget extend, this is cycle-level accounting
+            #self.list_extend(self.failed_set, uuids)
+            # Forget failed_set, this is cycle-level accounting
+            #self.put_list(self, self.failed_set, uuids)
+
+            if finshed:
```
hitz: "finished" is mis-spelled

src/snovault/elasticsearch/indexer.py
```diff
@@ -228,44 +531,152 @@ def update_objects(self, request, uuids, xmin, snapshot_id):
...
         return errors
 
-    def update_object(self, request, uuid, xmin):
+    def update_object(self, request, uuid, xmin, restart=False):
+        request.datastore = 'database'  # required by 2-step indexer
+
+        # If a restart occurred in the middle of indexing, this uuid might have already been indexd, so skip redoing it.
+        if restart:
+            try:
+                #if self.es.exists(index=self.index, id=str(uuid), version=xmin, version_type='external_gte'):  # couldn't get exists to work.
```
hitz: maybe it works in ES5?  Otherwise I guess I don't want this line here.

src/snovault/elasticsearch/indexer.py
```diff
@@ -46,118 +302,146 @@ def index(request):
     recovery = request.json.get('recovery', False)
     es = request.registry[ELASTIC_SEARCH]
     indexer = request.registry[INDEXER]
-
+    stage_for_followup = request.registry.settings.get("stage_for_followup", False)
```
hitz: What is this supposed to be for??  There are no tests, no config...

src/snovault/elasticsearch/indexer.py
tdreszer: This state object IS used and most of its methods are used in indexeer.py.  Those not are used in secondary_indexer.py.

src/snovault/elasticsearch/indexer.py
tdreszer: in encoded stage_for_followup is set, else secondary indexer never runs.  encoded test_indexer.py failed because the app didn't have that setting and I had to add it.  So this is tested in encoded.

PR #64, opened 2017-12-13
ENCD-3755-remove-schema

keenangraham:
This surface-level doc update related to same ticket in encoded repo.


hitz:
This doc change is not relevant to snovault -- the general case does not have these object types.

PR #67, opened 2018-01-25
Encd 3825 too many santa clauses

tdreszer:
This is the narrow try/except solution for the too many clauses indexing bug.  It also includes a version change on psycopg2, which fixes a travis build problem.  That change should not affect the encoded build at all.


caseylitton:
Cleaning up PR into two commits for master

PR #70, opened 2018-01-26
Encd 3825 increase max clauses count

tdreszer:
This should be the RC3.  It contains nax_clauses to 8K, no try/except, but an if len > 8K before the query.


caseylitton:
This was split into SNO-11 and SNO-12.  New PR's were submitted for both.

PR #71, opened 2018-01-30
Encd 3619 indexer v2 api (Tim's PR #1)

tdreszer:
This is the first of 4 expected PRs for several tickets that are interrelated.  The 3 others depend upon this code.  It introduces the indexer API /_indexer_state, which will give better indexer state information as well as enable requesting full reindex or for a small set of uuids.


caseylitton:
Combined with https://github.com/ENCODE-DCC/snovault/pull/72 in https://github.com/ENCODE-DCC/snovault/pull/79

PR #80, opened 2018-02-17
SNO-19 Update DOI preferred resolver url

katrinleinweber:
Hello!

The DOI foundation recommends a [new, secure resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8).

Cheers :-)


hitz:
this is dead code in snowflakes.  I am sure that it's fine to add if it was fine in encoded.

PR #85, opened 2018-02-27
SNO-25 Make uniqueItems to check the serialized values

davisagli:
This addresses https://encodedcc.atlassian.net/browse/SNO-25 where if the same linked object was specified in two different forms (i.e. uuid and @id), uniqueItems would not catch it.

I'm fixing this by overriding the `uniqueItems` validator to look at the validated, serialized values returned from the `items` subschema (in the case discussed in the ticket, those are normalized to uuids by the `linkTo` validator) rather than the raw values from the request. This will only be in place if the `items` validator is processed before the `uniqueItems` validator; fortunately the `jsonschema` library explicitly sets up `items` to run early.

### DIFF
--- a/src/snovault/schema_utils.py
+++ b/src/snovault/schema_utils.py
@@ -250,6 +250,17 @@ def permission(validator, permission, instance, schema):
         yield IgnoreUnchanged(error)
 
 
+orig_uniqueItems = Draft4Validator.VALIDATORS['uniqueItems']
+
+
+def uniqueItems(validator, uI, instance, schema):
+    # Use serialized items if available
+    # (this gives the linkTo validator a chance to normalize paths into uuids)
+    if validator._serialize and validator._validated[-1]:
+        instance = validator._validated[-1]
+    yield from orig_uniqueItems(validator, uI, instance, schema)
+
+
 VALIDATOR_REGISTRY = {}
 
 
@@ -279,6 +290,7 @@ class SchemaValidator(Draft4Validator):
     VALIDATORS['linkFrom'] = linkFrom
     VALIDATORS['permission'] = permission
     VALIDATORS['requestMethod'] = requestMethod
+    VALIDATORS['uniqueItems'] = uniqueItems
     VALIDATORS['validators'] = validators
     SERVER_DEFAULTS = SERVER_DEFAULTS
 
--- a/src/snovault/tests/test_schema_utils.py
+++ b/src/snovault/tests/test_schema_utils.py
@@ -1,0 +1,34 @@
+from snovault.schema_utils import validate
+import pytest
+
+
+targets = [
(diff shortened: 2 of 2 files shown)


davisagli:
@hitz @caseylitton this ticket is now marked as ready for PR

note that https://github.com/ENCODE-DCC/encoded/pull/2162 must be merged before `encoded` is updated to a `snovault` release that includes this.

src/snovault/schema_utils.py
```diff
@@ -250,6 +250,17 @@ def permission(validator, permission, instance, schema):
...
 
+orig_uniqueItems = Draft4Validator.VALIDATORS['uniqueItems']
+
+
+def uniqueItems(validator, uI, instance, schema):
+    # Use serialized items if available
+    # (this gives the linkTo validator a chance to normalize paths into uuids)
+    if validator._serialize and validator._validated[-1]:
+        instance = validator._validated[-1]
+    yield from orig_uniqueItems(validator, uI, instance, schema)
```
caseylitton: Why use 'from'?  I haven't read much about 'yield from' but it seems like yield would be fine.

In the BaseValidator code,
https://github.com/lrowe/jsonschema_serialize_fork/blob/master/jsonschema_serialize_fork/validators.py 
the iter_errors gets orig_uniqueItems as a generator on line 110.

Then it's looped over and yields the errors.  It's base code, so we can't change it, but shouldn't the 'yield from' be on 121? 

Now having looked into it, I wonder if we need a yield at all in the uniqueItems wrapper?  Should we yield the orig_uniqueItems generator function or just return it directly?

src/snovault/schema_utils.py
```diff
@@ -250,6 +250,17 @@ def permission(validator, permission, instance, schema):
         yield IgnoreUnchanged(error)
 
 
+orig_uniqueItems = Draft4Validator.VALIDATORS['uniqueItems']
```
caseylitton: I would put this up top.  Is there a reason to do it here?

src/snovault/schema_utils.py
davisagli: `yield from` lets a generator to delegate to another generator (see PEP 380). So this line is equivalent to:

```python
for error in orig_uniqueItems(validator, uI, instance, schema):
    yield error
```

I hadn't looked at how validators are executed, so was assuming it is important for the `uniqueItems` validator to continue to be a generator. As it turns out, it looks like it would probably be okay to just do `return orig_uniqueItems(validator, uI, instance, schema)` since `iter_errors` only assumes the return value of a validator is iterable, not that it is necessarily a generator. I'd still feel more comfortable keeping it as a generator since I haven't done a thorough hunt to see if there are other call sites.

src/snovault/schema_utils.py
davisagli: If I'm only using something once I prefer to keep its definition close to where it's used.

PR #86, opened 2018-03-01
SNO-26 Add schemas map view to profiles

forresttanaka:
PRing so Laurence can review to see if I’ve gone mad. 😜

The main question is whether adding this new /profiles-map/ route is a good idea. When viewing a schema now, we can have `<a>` links to other schemas via the linkTo and linkFrom properties. But those only contain the other schema’s `@type` that doesn’t have a necessarily obvious mapping to a schema path. I could have done a GET on /profiles/ but that returns 1.2MB JSON. So I made this new path that has the same dict keys as /profiles/ but the values are just the schema path strings instead of the entire string object, and right now that returns in 5.9KB JSON.

```
{
    "AccessKey": "/profiles/access_key_admin.json",
    "AnalysisStep": "/profiles/analysis_step.json",
    "AnalysisStepRun": "/profiles/analysis_step_run.json",
    "AnalysisStepVersion": "/profiles/analysis_step_version.json",
    "Annotation": "/profiles/annotation.json",
...
```

### DIFF
--- a/src/snovault/schema_views.py
+++ b/src/snovault/schema_views.py
@@ -11,6 +11,8 @@ from .util import mutated_schema
 def includeme(config):
     config.add_route('schemas', '/profiles/')
     config.add_route('schema', '/profiles/{type_name}.json')
+    config.add_route('schemap', '/profiles/{type_name}')
+    config.add_route('schemas_map', '/profiles-map/')
     config.scan(__name__)
 
 
@@ -39,6 +41,8 @@ def _annotated_schema(type_info, request):
 
 @view_config(route_name='schema', request_method='GET',
              decorator=etag_app_version_effective_principals)
+@view_config(route_name='schemap', request_method='GET',
+             decorator=etag_app_version_effective_principals)
 def schema(context, request):
     type_name = request.matchdict['type_name']
     types = request.registry[TYPES]
@@ -65,3 +69,15 @@ def schemas(context, request):
         subtypes[name] = type_info.subtypes
 
     return schemas
+
+
+@view_config(route_name='schemas_map', request_method='GET',
+             decorator=etag_app_version_effective_principals)
+def schemas_map(context, request):
+    types = request.registry[TYPES]
+    profiles_map = {}
+    for type_info in types.by_item_type.values():
+        if 'id' in type_info.schema:
+            profiles_map[type_info.name] = type_info.schema['id']
+    profiles_map['@type'] = ['JSONSchemas']
+    return profiles_map


lrowe:
With the benefit of hindsight, I should probably have made /profiles/ a JSONSchemaCollection so that listings worked the same way for the schemas as they do for collections of regular types (though that might require indexing them in Elasticsearch to work.)

From a JSON-LD perspective, I think you can avoid creating a new 'JSONSchemasMap' type since what it returns is valid under the existing 'JSONSchemas' type used for /profiles/. The json response for /profiles/ is a little like `?frame=expand` while your new /profiles-map/ is like `?frame=object`. (Whether a JSON-LD property value is a url string pointing to the object or that object inlined, they're semantically equivalent.)

While it would be fine from a JSON-LD perspective to just make /profiles/ return without the individual JSONSchema objects inlined (and only expand them all out for `?frame=expand`), I think some submitter labs at one point relied on fetching the /profiles/ to power some of their own custom forms.

src/snovault/schema_views.py
```diff
@@ -65,3 +69,30 @@ def schemas(context, request):
...
+
+    To direct this response to a front-end React component, the returned JSON
+    also includes an @type of 'JSONSchemasMap'.
+    """
+    types = request.registry[TYPES]
+    map = {}
+    for type_info in types.by_item_type.values():
+        if 'id' in type_info.schema:
+            map[type_info.name] = type_info.schema['id']
+
```
caseylitton: remove extra spaces

src/snovault/schema_views.py
```diff
@@ -65,3 +69,30 @@ def schemas(context, request):
...
+    map['@type'] = ['JSONSchemas']
+
```
caseylitton: remove extra spaces

src/snovault/schema_views.py
```diff
@@ -65,3 +69,30 @@ def schemas(context, request):
...
+    only assembles and returns a dict keyed by each schema object name (e.g.
+    "GeneticModification") with a value of the corresponding schema ID (in this
+    example that would be "/profiles/genetic_modification.json"). Any schemas
+    without an `id` property don't get included.
+
+    To direct this response to a front-end React component, the returned JSON
+    also includes an @type of 'JSONSchemasMap'.
+    """
+    types = request.registry[TYPES]
+    map = {}
```
caseylitton: map is a built in function so we can't use it as a variable.

src/snovault/schema_views.py
```diff
@@ -65,3 +69,30 @@ def schemas(context, request):
         subtypes[name] = type_info.subtypes
 
     return schemas
+
+
+@view_config(route_name='schemas_map', request_method='GET',
+             decorator=etag_app_version_effective_principals)
+def schemas_map(context, request):
+    """Returns a mapping of schema object name to the corresponding schema ID.
```
caseylitton: No need for comments like these in the python code.  We don't do that.  Please remove.  

Effort should be directed at writing python code that doesn't need comments.  Which I think the function does fine.  Comments end up repeating what someone could glean from just looking at the code.

davisagli:
> No need for spaces in functions unless they are really big. But then I would say the function should be split up. Spaces go between functions in python.

I wouldn't be dogmatic about this. There are times when the most readable code is a single function that does several groups of things in sequence with blank lines to help indicate the grouping.

forresttanaka:
Are we following any style guide for Python? Someone asked me (@caseylitton maybe?) that a few months ago and I didn’t know, and our existing code seems to be a mix, as expected. Since Nathan said he followed Google’s, I’ve been following that as best I can without memorizing the entire thing.

https://google.github.io/styleguide/pyguide.html

So let me reply to each with a section from the guide as appropriate, unless there‘s another guide we’re following that I don’t know about.

src/snovault/schema_views.py
forresttanaka: OK; didn’t know about that.

src/snovault/schema_views.py
forresttanaka: https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

“A function must have a docstring, unless it meets all of the following criteria:”

I think this function meets the second, but not the other two.

src/snovault/schema_views.py
forresttanaka: https://google.github.io/styleguide/pyguide.html?showone=Blank_Lines#Blank_Lines

“Use single blank lines as you judge appropriate within functions or methods.”

davisagli:
Huh, I didn't know about that google styleguide and will need to take a look. AFAIK the standard guide in the Python world is [PEP8](https://www.python.org/dev/peps/pep-0008/).

forresttanaka:
Oh; PEP8 sounds more authoritative (well, at least their web page is formatted better than Google’s). Comparing the sections, they don’t seem _too_ different. PEP8 adds “sparingly” about blank lines, but puts more emphasis on function docstrings or comments.

davisagli:
There's also https://www.python.org/dev/peps/pep-0257/ specifically about how to format docstrings.

src/snovault/schema_views.py
caseylitton: I agree.   Spaces after indented code blocks and before returns are inappropriate.   They don't add anything to the code.

forresttanaka:
I’ll make these changes, but right now every Python PR is going to feel like a pot shot in the dark until we get there.

PR #87, opened 2018-03-06
SNO-29 Limit audits on form update

davisagli:
This makes 2 changes to help performance just after an item is edited:
1. If an item is fetched from elasticsearch but turns out to have been invalidated, then set the request's datastore to `database` so that fetches of other items will go directly to the database (basically, assume that they are likely to have been invalidated too).
2. Don't embed audits in items unless they were fetched from elasticsearch. This way we only incur the overhead of calculating audits while indexing.

### DIFF
--- a/src/snovault/elasticsearch/esstorage.py
+++ b/src/snovault/elasticsearch/esstorage.py
@@ -80,11 +80,17 @@ class PickStorage(object):
             return self.read
         return self.write
 
+    def force_database(self):
+        request = get_current_request()
+        if request:
+            request.datastore = 'database'
+
     def get_by_uuid(self, uuid):
         storage = self.storage()
         model = storage.get_by_uuid(uuid)
         if storage is self.read:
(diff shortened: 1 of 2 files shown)


src/snovault/elasticsearch/esstorage.py
```diff
@@ -80,11 +80,17 @@ def storage(self):
             return self.read
         return self.write
 
+    def force_database(self):
```
caseylitton: We're not using the self reference.   @staticmethod should be applied and the self removed.

davisagli: Fair point, but I just turned it into a function at module scope since there's no particular reason it belongs in the PickStorage class.

davisagli:
@caseylitton this is done once you and Ben review my latest changes; there won't be a PR for ENCD-3885 from me because the only change that's needed is updating to a release of snovault that includes this PR.

PR #88, opened 2018-03-07
Rebase/sno 09 10 indexer updates and reverts

caseylitton:
Rebased version of #79


caseylitton:
Fixed the origin PR #79

PR #93, opened 2018-03-22
SNO-28 limit ES storage to indices created for snovault resources

davisagli:
This avoids searching unrelated indexes such as the vis cache and regions. Seems to make getting a single object 40-100% faster on my laptop with test data.


davisagli:
@hitz need anything from me to finish reviewing this?

PR #94, opened 2018-03-28
SNO-33 Specify index for get_by_unique_key from collection

davisagli:
This will hopefully improve performance for fetching single objects at their canonical URI.

This is based on the changes from #93 so you can check the cumulative performance improvement in a demo -- start one from the SNO-33-one-index branch of `encoded`, which uses this snovault branch. Look at the more recent commit to see just the changes related to SNO-33.

Note that this involves a small change to the interface for storages -- the `get_by_unique_key` method must now take an optional `index` parameter.

### DIFF
--- a/src/snovault/resources.py
+++ b/src/snovault/resources.py
@@ -168,7 +168,12 @@ class AbstractCollection(Resource, Mapping):
                 return default
             return resource
         if self.unique_key is not None:
-            resource = self.connection.get_by_unique_key(self.unique_key, name)
+            # Give the storage a hint of which index to search.
+            # If this is a collection of an abstract type,
+            # the item_type will be None and we'll search all snovault indices.
+            index = getattr(self.type_info, 'item_type', None)
+            resource = self.connection.get_by_unique_key(
+                self.unique_key, name, index=index)
             if resource is not None:
                 if not self._allow_contained(resource):
                     return default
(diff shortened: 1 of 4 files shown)


src/snovault/resources.py
```diff
@@ -168,7 +168,12 @@ def get(self, name, default=None):
                 return default
             return resource
         if self.unique_key is not None:
-            resource = self.connection.get_by_unique_key(self.unique_key, name)
+            # Give the storage a hint of which index to search.
+            # If this is a collection of an abstract type,
```
hitz: Now, THAT I did not consider.   We would have to register abstract types more firmly if we were to map an abstract class to all it's subclasses.   I'll think about this as a possible enhancement.

davisagli: Given that it's a performance optimization and looking up specific items within an abstract collection is probably uncommon, I doubt it's very high priority.

davisagli:
> I would like to see a define something like: SNOVAULT_RESOURCES = 'snovault-resources' .

Yeah, good call. Will do.

> Is 'meta' excluded from snovault-resources? It should be.

Good catch. Will fix.

> A bit beyond this change, but some indices should get grouped by alias (e.g. DATASET_INDEXES = 'experiment,annotation,...').

That sounds like a separate, `encoded`-specific feature request.

davisagli:
Updated.

PR #104, opened 2018-05-10
SNO-36-update-buildout

keenangraham:
Update buildout procedure to be same as ENCODED.


versions.cfg
```diff
@@ -28,9 +28,9 @@ rutter = 0.2
 selenium = 3.4.3
 splinter = 0.7.5
 translationstring = 1.3
-urllib3 = 1.16
+urllib3 = 1.14
```
davisagli: Hmm, a downgrade of urllib3?

keenangraham: Wasn't sure why it wasn't working before I removed `-index = https://pypi.python.org/simple/` so aligned it with versions in encoded: https://github.com/ENCODE-DCC/encoded/blob/ab5124cba81b7d0840cd6eb68d00ebd622726068/versions.cfg#L30

Alternatively we could update the one in encoded.

PR #106, opened 2018-05-14
SNO-35-mismatched-storage-objects

keenangraham:
Removing the global context switch between database/elasticsearch. Integrity of the system outweighs performance gain.


keenangraham:
Too slow. Closing for now.

PR #110, opened 2018-05-30
SNO-40-upgrade-boto3

keenangraham:
This tries to recreate the boto functionality in S3BlobStorage with boto3. Several boto methods don't have equivalent methods in boto3. The storage parameters `read_profile_name=None, store_profile_name=None` never seemed to be used so removed. The `store_conn` and `read_conn` abstractions likewise didn't seem useful to recreate.

### DIFF
--- a/src/snovault/storage.py
+++ b/src/snovault/storage.py
@@ -49,9 +49,7 @@ def includeme(config):
     _DBSESSION = registry[DBSESSION]
     if registry.settings.get('blob_bucket'):
         registry[BLOBS] = S3BlobStorage(
-            registry.settings['blob_bucket'],
-            read_profile_name=registry.settings.get('blob_read_profile_name'),
-            store_profile_name=registry.settings.get('blob_store_profile_name'),
+            registry.settings['blob_bucket']
         )
     else:
         registry[BLOBS] = RDBBlobStorage(registry[DBSESSION])
--- a/src/snovault/storage.py
+++ b/src/snovault/storage.py
@@ -49,7 +49,9 @@ def includeme(config):
     _DBSESSION = registry[DBSESSION]
     if registry.settings.get('blob_bucket'):
         registry[BLOBS] = S3BlobStorage(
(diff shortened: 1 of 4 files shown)


src/snovault/tests/test_storage.py
```diff
@@ -223,45 +223,50 @@ def test_keys(session):
...
 
 
-def test_S3BlobStorage(mocker):
+def test_S3BlobStorage_boto3(mocker):
     from snovault.storage import S3BlobStorage
-    mocker.patch('boto.connect_s3')
+    mocker.patch('boto3.resource')
     bucket = 'test'
-    fake_key = mocker.Mock()
+    mock_key = mocker.Mock()
```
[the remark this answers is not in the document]
keenangraham: Looks neat.

src/snovault/storage.py
```diff
@@ -48,9 +49,7 @@ def includeme(config):
     _DBSESSION = registry[DBSESSION]
     if registry.settings.get('blob_bucket'):
         registry[BLOBS] = S3BlobStorage(
-            registry.settings['blob_bucket'],
-            read_profile_name=registry.settings.get('blob_read_profile_name'),
```
caseylitton: Summarizing our conversation about the ini/cfg/s3 aws profile vars and usage.

We should reconsider removing the 'blob_read_profile_name' and 'blob_store_profile_name'.  Or clean up buildout config/ini files.  The current PR seems to work because boto just uses the default profile.  Removing them here leaves unused code in the ini/cfgs and limits our ability to set the profile.  

We could do nothing, which would limit our profile functionality and leave some technical debt in the ini/configuration.   Or we could keep the profile functionality by updating the PR.   Cleaning up the configuration would be a big task.  It would probably require discussion about what functionality we want/need and who should have permission on demo/production.

keenangraham: Probably should keep profile_names in case someone else is using them (even though it doesn't seem to make a difference for us). Will update PR.

PR #91, opened 2018-03-21
SNO-31-32 Create search helpers and lint search

kakaly:
**SNO 31 create helpers**

- Adds helpers/helper.py module

- Removes helper methods from search.py

- Resolves dependencies

- Fixes lint errors

PR #114, opened 2018-06-27
ENCD-3692 Snovault search refactor 

kakaly:
This change has corresponding encoded changes here: https://github.com/ENCODE-DCC/encoded/pull/2319


src/snovault/vis_defines.py
```diff
@@ -0,0 +1,1614 @@
...
+        #                },
+        #                { ... }
+        #            ],
+        #            "signal_reverse": [ { ... } ]
+        #        }
+        #    },
+        #    "experiment_2": {
+        #        ...
+        #    },
+
```
caseylitton: Spoke to Karthik about adding the vis_defines.py file.  This is an unfortunate dependency in helpers.py on vis_format_url without a clear solution.  We'll want to examine this in more detail.  Probably resolve/merge that ticket, ENCD-4116, first and then integrate it with search.  It would be a bad idea to hide that change in the search update ticket. 

notes:
-Copying the whole file into snovault is redundant and confusing, but the easiest.  
-Moving the whole file to snovault would split the visualization code.    
-We definitely don't want snovault dependent on encoded either.

zoldello: @caseylitton I addressed ENCD-4116, a while ago. I recommend thinking about deploying that after this current PR is settled.

zoldello:
This change (114) subsumed 91 (https://github.com/ENCODE-DCC/snovault/pull/91)

zoldello:
This serves as a reference. It has been closed in favor of https://github.com/ENCODE-DCC/snovault/pull/120.

PR #116, opened 2018-07-05
SNO-43 Clean up snovault startup

kilodalton:
I found these minor modifications to be necessary to start up the snovault repo "as-is".

This repo and this PR does not fully match the current version of [encoded](https://github.com/ENCODE-DCC/encoded/) in terms of JS handling and CSS building (e.g. encoded uses `webpack` to build CSS now) but this was the smallest set of changes to get `snovault` running locally.

Importantly the rubygems recipe was removed, and the ruby gem is expected to be manually installed locally, and the ruby `gem install compass` has been added to the TravisCI config.

---
**:warning: Note:**
There are some outdated npm packages which can create a problem during install (buildout stops):
`Error: Cannot find module 'internal/util/types'` 

Trying this sequence of commands may or may not help (the problem seems to be widespread in the community because of some particular dependencies, but the exact steps to fix it are hard to come by... other than updating the older/deprecated npm packages causing the problems of which there are a few still in this repo):

```bash
rm -r node_modules/*
npm cache clear --force
npm install
```

and possibly (if you used homebrew to install node):
```bash
/usr/local/bin/npm install npm@latest -g
```

Most likely candidates to evaluate for upgrade because of their use of an older version of `graceful-fs`:
- `gulp`


kilodalton:
The TravisCI tests are failing for reasons that are unlikely to be related to my changes but rather to the inability to pull in the commit-specific versions of `react-forms` and `whatwg-fetch`, e.g.:
>Could not install /home/travis/build/ENCODE-DCC/snovault/react-forms@git+https:/github.com/lrowe/react-forms.git#3953a633b1f77640dffb5e3f1d5bbe78a98c3dfe

image

versions.cfg
```diff
@@ -28,7 +28,7 @@ rutter = 0.2
 selenium = 3.4.3
 splinter = 0.7.5
 translationstring = 1.3
-urllib3 = 1.14
+urllib3 = 1.22
```
hitz: this is probably what is blocking local snovault builds.  Checking now.

hitz:
@kilodalton what version of botocore are you using?  When I try to start dev-servers on the mac I get `pkg_resources.ContextualVersionConflict: (botocore 1.9.11 (/Users/hitz/.pyenv/versions/3.4.3/lib/python3.4/site-packages), Requirement.parse('botocore<1.6.0,>=1.5.0'), {'boto3'})`

encodeD is current using 1.10.30... I am going to try to bump it.

kilodalton:
Update regarding the comment above, it has been confirmed by myself and @hitz  that this branch can build/test/run the repo as-is (with this branch) without errors using Python 3.4.

I specifically tested:
 - Python 3.4.3 ✅ 
 - Python 3.4.8 ✅

kilodalton:
@caseylitton 

@hitz had asked me to make this PR so that folks using `snovault` would be able to spin in it up more easily. Do you think you will merge it? It should be non-impactful on `encoded` function.

PR #120, opened 2018-07-24
SNO-31-search-refactor

zoldello:
This was copied off: https://github.com/ENCODE-DCC/snovault/tree/snovault-refactor-low-lvl. That branch now serves as a reference to Kathik's work. All further work should be done off this branch.


src/snovault/viewconfigs/searchview.py
```diff
@@ -0,0 +1,320 @@
...
+            self.request.response.status_code = 404
+            self.result['notification'] = 'No results found'
+            self.result['@graph'] = []
+        return total
+
+    def preprocess_view(self):
+        self.result['@id'] = '/search/' + self.search_base
+        self.result['@type'] = ['Search']
+        self.result['title'] = 'Search'
+        self.result['clear_filters'] = self.construct_clear_filters()
```
caseylitton: construct_clear_filters is dependent on doc_types being build first.  Is looks like the current implementation builds the clear filters when self.doc_types = [].  

Moving self.construct_doc_types() above self.result['clear_filters'] = self.construct_clear_filters() is NOT the  solution.  The bottom code block in construct doc types needs to be below construct_clear_filters.  See that comment above.  The code needs to be untied;  We can't cleanly construct one before the other.

src/snovault/viewconfigs/searchview.py
```diff
@@ -0,0 +1,320 @@
...
+
+        try:
+            self.doc_types = sorted({self.types[name].name for name in self.doc_types})
+        except KeyError:
+            # Check for invalid types
+            bad_types = [t for t in self.doc_types if t not in self.types]
+            msg = "Invalid type: {}".format(', '.join(bad_types))
+            raise HTTPBadRequest(explanation=msg)
+
+        if not self.doc_types:
```
caseylitton: See other comment in preprocess.  This conditional needs to happen after construct_clear_filters.  And the code above this line in construct_doc_types needs to happen before construct_clear_filters

src/snovault/viewconfigs/searchview.py
```diff
@@ -0,0 +1,320 @@
...
+    def preprocess_view(self):
+        self.result['@id'] = '/search/' + self.search_base
+        self.result['@type'] = ['Search']
+        self.result['title'] = 'Search'
+        self.result['clear_filters'] = self.construct_clear_filters()
+
+        # Construct self.doc_types
+        self.construct_doc_types()
+
+        if self.doc_types != ['Item'] and self.doc_types != DEFAULT_DOC_TYPES:
```
caseylitton: This is related to the code bock we need to move out of contruct_doc_types.  That block goes here and determines if we build filters and views into the result.

src/snovault/viewconfigs/searchview.py
```diff
@@ -0,0 +1,320 @@
...
+                self.doc_types = ['Item']
+            # For /search/ with no type= use defalts
+            else:
+                self.doc_types = DEFAULT_DOC_TYPES
+
+    def construct_result_views(self):
+
+        # Add special views like Report and Matrix if search is a single type
+        for item_type in self.doc_types:
+            ti = self.types[item_type]
```
caseylitton: This is original code but it messy.  'ti' is set in the for loop then used to in adding views as ti.factory.

This line is equivalent to ti = self.types[-1].  if one exists, which it should at the point give how doc_types is set.  Since it's a factory this probably doesn't matter but then there would be no need to loop through the doc types.  Just set 'ti = self.types[-1]'.  I suggest checking the outcome of ti.factory on different doctypes.  Probably has zero impact.

src/snovault/viewconfigs/searchview.py
zoldello: doc_type is initialized in BaseView (base_view.py, line 24, code- "self.doc_types = self.request.params.getall('type')"). Where is self.doc_types set as an empty list?

PR #126, opened 2018-08-07
SNO-47 Add index time log option and update index output data

caseylitton:
This PR updates the Indexer.update_object function so it returns more information to update_objects.  This is the only way to get info back from update_object when running MPIndexer.update_objects since it's in sub processes.

Also this PR adds an index_logger class that is imported and added to indexer.py:Indexer class.  The logger is added to all indexers, viz, region, and primary but is only turned on in the primary indexer.  So includeme was updated in indexer.py and mpindexer.py.

I will add this to the ini configs so we can turn it on/off without changing repo code.

### DIFF
--- a/src/snovault/elasticsearch/index_logger.py
+++ b/src/snovault/elasticsearch/index_logger.py
@@ -63,7 +63,7 @@ class IndexLogger(object):
         if 'es_time' in output and isinstance(output['es_time'], float):
             output['es_time'] = '%0.6f' % output['es_time']
         self.write_log(
-            '{timestamp} {doc_path} '
+            '{timestamp} {doc_path} {doc_type} '
             '{embed_time} {embed_ecp} '
             '{es_time} {es_ecp} '
             '{embeds} {linked} '
@@ -75,6 +75,7 @@ class IndexLogger(object):
                 es_ecp=output.get('es_ecp'),
                 doc_path=output.get('doc_path'),
                 linked=output.get('doc_linked'),
+                doc_type=output.get('doc_type'),
                 timestamp=output.get('timestamp'),
             )
         )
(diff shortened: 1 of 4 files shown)


src/snovault/elasticsearch/index_logger.py
```diff
@@ -0,0 +1,98 @@
...
+        '''Close all log handlers'''
+        for handler in self._the_log.handlers:
+            handler.close()
+            self._the_log.removeHandler(handler)
+
+    def _get_log(self):
+        if self._do_log:
+            file_name = "{}-{}.log".format(
+                self.log_name,
+                str(int(time.time() * 10000000)),
```
hitz: this is converting microsec-> sec?   Probably should have a comment here.

src/snovault/elasticsearch/index_logger.py
```diff
@@ -0,0 +1,98 @@
...
+            return log
+        return None
+
+    def _reset_log(self):
+        '''
+        Close handlers and Get new log
+
+        * Logger gets logs name so we call twice to clear and get a new one
+        '''
+        if self._do_log:
```
hitz: this seems super clunky but I guess you have a good reason?

caseylitton: eh...It's clunky.  I didn't like it.  I used the flag in the logger class to limit the indexer code changes.  Now that it works maybe I can find a better solution.

src/snovault/elasticsearch/index_logger.py
caseylitton: Well yes and no, It's just a way to keep timestamps on the file.  I can add a comment anyhow.

caseylitton:
Things to update, 

add time.time() for end time
add pid to log
backoff time
look through embedd stack

caseylitton:
reopen in future

PR #131, opened 2018-08-30
SNO-52-alter-select-distinct-values

keenangraham:
Checks to see if field is actually calculated property before calculating it on @@object in select_distinct_value function.

### DIFF
--- a/src/snovault/util.py
+++ b/src/snovault/util.py
@@ -60,6 +60,7 @@ def expand_path(request, obj, path):
 def _get_calculated_properties_from_paths(request, paths):
     root = request.registry[ROOT]
     calculated_fields = set()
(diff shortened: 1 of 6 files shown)


src/snovault/util.py
```diff
@@ -56,16 +57,32 @@ def expand_path(request, obj, path):
         expand_path(request, value, remaining)
 
 
+def get_calculated_properties_from_paths(request, paths):
+    root = request.registry[ROOT]
+    calculated_fields = set()
+    item_types = {
+        root.collections.get(p.split('/')[1]).type_info.item_type
```
hitz: don't like naked [1] here.  I think it's better to add a line of code parsing the split()

keenangraham:
Added tests for skip_calculated on object view, get_calculated_properties_from_paths, and new select_distinct_values function.

PR #134, opened 2018-09-07
SNO-51 Clean up indexer initialization

caseylitton:
Goes with ENCD-4198 PR


caseylitton:
Not necessary before es worker

PR #137, opened 2018-09-11
SNO-53+59 Add index flags to indexers+Log name

caseylitton:
Goes with ENCD-4209 PR.  Someone could add a reminder to revert the short indexer commit to their approval.  I would but I can't review my own pr.

There are three indexer states flagged in the PR.  
1.  The initial indexing state.  First index or indexing has been reset.
2. Reindex = all.  This end point resets the indexer.  This case and reindex and init index flags as true.
3. Reindex not all.  This end pint creates a list of uuids to reindex and does not reset the indexer.

After this PR I will submit a indexing logger PR, SNO-47.  It will dump indexer data to a json file when one or both of these variables are true.

### DIFF
--- a/src/snovault/elasticsearch/indexer.py
+++ b/src/snovault/elasticsearch/indexer.py
@@ -137,6 +137,7 @@ def index(request):
     state = IndexerState(es, INDEX, followups=stage_for_followup)
 
     (xmin, invalidated, restart) = state.priority_cycle(request)
+    state.log_reindex_init_state()
     # OPTIONAL: restart support
     if restart:  # Currently not bothering with restart!!!
         xmin = -1
--- a/src/snovault/elasticsearch/indexer.py
+++ b/src/snovault/elasticsearch/indexer.py
@@ -35,7 +35,7 @@ import requests
 
 es_logger = logging.getLogger("elasticsearch")
 es_logger.setLevel(logging.ERROR)
-log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


src/snovault/elasticsearch/indexer.py
```diff
@@ -228,6 +229,12 @@ def index(request):
                     snapshot_id = connection.execute('SELECT pg_export_snapshot();').scalar()
 
     if invalidated and not dry_run:
+        invalid = []
+        for uuid in invalidated:
+            if len(invalid) < 100:
```
keenangraham: Must remove before merging I think.

keenangraham:
Possible to add some tests that call the indexer_state endpoint?

src/snovault/elasticsearch/indexer.py
keenangraham: Must remove before merging I think.

caseylitton: yep,  I'd flag it myself but I can't review my own pr.

caseylitton:
@hitz yea, I'm not a fan of the getter/setter thing either.  In this case, the way the indexer_state is being used as a wrapper around elastic_search meta data with hard coded string makes me uncomfortable.  Then slack is integrated into there too with the notifications.  There are three things going on, es wrapper, indexer state management, and slack notifications...

PR #146, opened 2018-10-16
SNO-57 SNO-51 cleanup indexer

caseylitton:
Goes with ENCD-4198

This also has SNO-53/57, but that should be merged in v76.  This is going for v77


caseylitton:
reopen in future

PR #147, opened 2018-10-17
SNO-60-check-if-res-has-view

zoldello:


src/snovault/viewconfigs/report.py
```diff
@@ -53,6 +53,9 @@ def preprocess_view(self, views=None, search_result_actions=None):
             views=views,
             search_result_actions=search_result_actions,
         )
+        #TODO: Investigate res['views'] is not always set
+        if 'views' not in res:
```
caseylitton: Let's make this a little more robust.  The current update doesn't handle the case of views existing as an empty list.
```
override_view = {
    'href': res['@id'],
    'title': 'View results as list',
    'icon': 'list-alt',
}
res_views = res.get('views')
if not res_views:
    res['views'] = [override_view]
else:
    res['views'][0] = override_view
```

zoldello: I tweaked  your code a little. Please, let me know if what I did is acceptable.

PR #151, opened 2018-11-02
SNO-65 Refactor index UUID server

caseylitton:
https://github.com/ENCODE-DCC/encoded/pull/2485

### DIFF
--- a/src/snovault/elasticsearch/indexer.py
+++ b/src/snovault/elasticsearch/indexer.py
@@ -5,6 +5,7 @@ from elasticsearch.exceptions import (
     TransportError,
 )
 from pyramid.view import view_config
+from pyramid.settings import asbool
 from sqlalchemy.exc import StatementError
 from snovault import (
     COLLECTIONS,
@@ -25,6 +26,11 @@ from .indexer_state import (
     all_types,
     SEARCH_MAX
 )
+from .simple_queue import (
+   SimpleUuidServer,
+   SimpleUuidClient,
+)
+
 import datetime
 import logging
 import pytz
@@ -237,7 +243,13 @@ def index(request):
 
         # Do the work...
 
-        errors = indexer.update_objects(request, invalidated, xmin, snapshot_id, restart)
+        errors = indexer.serve_objects(
+            request,
+            invalidated,
+            xmin,
+            snapshot_id=snapshot_id,
+            restart=restart,
+        )
 
         result = state.finish_cycle(result,errors)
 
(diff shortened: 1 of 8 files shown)


keenangraham:
Restarting tests.

src/snovault/elasticsearch/indexer.py
```diff
@@ -293,13 +305,77 @@ def get_current_xmin(request):
...
+
+        )
+        queue_client_chunk_size = int(
+            registry.settings.get('queue_client_chunk_size', 1024)
+        )
+        queue_client_batch_size = int(
+            registry.settings.get('queue_client_batch_size', 1024)
+        )
+        if is_queue_server:
+            if not queue_type or queue_type == 'Simple':
```
keenangraham: What happens if there is a queue_type that's not `Simple`?

src/snovault/elasticsearch/indexer.py
```diff
@@ -293,13 +305,77 @@ def get_current_xmin(request):
...
+                self.queue_server = SimpleUuidServer()
+                if is_queue_client:
```
keenangraham: All queue_clients are queue_servers but not the other way around?

src/snovault/elasticsearch/indexer.py
caseylitton: Yes, it wouldn't work.  This is temporary.  

In the next PR we'll add a redis queue.  Then 'Simple' queue will be the default.  If, for example, the queue in the config doesn't exists or the config says 'Redis' but redis isn't installed or running. 

Other queues could be AWS SQS or even an ES queue.  But I only have a redis one functioning.  The goal is that this pattern will allow us to plug in other queues as needed.

src/snovault/elasticsearch/indexer.py
caseylitton: At some point I had the queue client inheriting from the queue server...or the other way around.  But then I realized that we'll need the queue server to optionally run it's own indexing.   I accomplished this by giving the queue server a client.  Then the serve_queue function the QueueServer checks if it has a client and then will run the worker process with it.

I'm not thrilled about it.  Kinda hacked it in late in the game.

src/snovault/elasticsearch/indexer.py
caseylitton: An additional thing here is that all processes try to add an indexer to their registry.  But only the primary, viz, and region indexers should, not the 6 encoded application process.  I have a fix for that in the indexer.py includeme in a different PR, so this code is a little off.

src/snovault/elasticsearch/simple_queue.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+class SimpleUuidServer(object):
+    '''Simple uuid queue as a list'''
+    def __init__(self):
+        self._uuids = []
+        self._client_conns = {}
+        self.errors = []
+
+    def _get_uuid(self):
```
zoldello: I think this technical should be a property object (https://www.programiz.com/python-programming/property)

src/snovault/elasticsearch/simple_queue.py
```diff
@@ -0,0 +1,156 @@
...
+                if not uuid:
+                    break
+                uuids.append(uuid)
+                cnt -= 1
+            client_conn['uuid_cnt'] = len(uuids)
+        return uuids
+
+    def has_uuids(self):
+        '''Are there uuids in the queue'''
+        if self._uuids:
```
zoldello: how about:

return  self._uuids

src/snovault/elasticsearch/simple_queue.py
```diff
@@ -0,0 +1,156 @@
...
+            return True
+        return False
+
+    def is_indexing(self):
+        '''Is an indexing process currently running'''
+        if self.has_uuids() or self.errors:
+            return True
+        for client_info in self._client_conns.values():
+            if client_info['uuid_cnt']:
+                return True
```
zoldello: How about: 

return false or client_info['uuid_cnt'] or not client_info['results']

src/snovault/elasticsearch/simple_queue.py
```diff
@@ -0,0 +1,156 @@
...
+    def load_uuids(self, uuids):
+        '''
+        The only way to load uuids in queue
+        - Uuids are to be loaded once per indexing session
+        '''
+        if self.has_uuids():
+            return None
+        for uuid in uuids:
+            self._load_uuid(uuid)
+        return len(self._uuids)
```
zoldello: Why is a load-method returning length? Why is it not throwing an exception if it cannot load?

src/snovault/elasticsearch/simple_queue.py
```diff
@@ -0,0 +1,156 @@
...
+            client_conn = self._client_conns[client_id]
+            client_conn['results'] = results
+            batch_errors = []
+            for error in results['errors']:
+                error['client_id'] = client_id
+                batch_errors.append(error)
+            self.errors.extend(batch_errors)
+            client_conn['uuid_cnt'] -= results['successes']
+            client_conn['uuid_cnt'] -= len(batch_errors)
+            if results['successes'] < 0 or client_conn['uuid_cnt']:
```
zoldello: ?:

msg = 'Queue Server cannot close client(%s)' % client_id
return  msg if results['successes'] < 0 or client_conn['uuid_cnt'] else 'Okay'

src/snovault/elasticsearch/tests/test_simple_queue.py
```diff
@@ -0,0 +1,416 @@
+''''
+Test the functionality of SimpleUuidServer and SimpleUuidClient
+in the context of Indexer and MPIndexer
+'''
+from os import getpid
+from unittest import (
```
zoldello: Why not pytest?

src/snovault/elasticsearch/simple_queue.py
```diff
@@ -0,0 +1,156 @@
...
+        if self.has_uuids() or self.errors:
+            return True
+        for client_info in self._client_conns.values():
+            if client_info['uuid_cnt']:
+                return True
+            elif client_info['results'] is None:
+                return True
+        return False
+
+    def load_uuids(self, uuids):
```
hitz: I think this method should be adjacent to _load_uuid()

src/snovault/elasticsearch/simple_queue.py
caseylitton: Possibly, these functions are 'stubs' for the more complex queue types, like redis or aws sqs.   It will be good to review this after seeing how the other queue types plug in.

src/snovault/elasticsearch/simple_queue.py
caseylitton: That would actually return the uuids.  Here we just want a way to verify there are uuids.  

Other queue types have to make requests to get queue info and have different ways to check the queue length.   AWS SQS queue won't actually return the exact queue len for example.  It estimates in a way.  And redis has different function call depending on the queue type, set or list...there are a few others.  

In the end we just want the has_uuids function to return True or False if there are uuids remaining.

src/snovault/elasticsearch/simple_queue.py
caseylitton: Okay, I'll hold this for the next pr when we plug in the remote queues.  There are a few functions set up with a similar pattern.

src/snovault/elasticsearch/simple_queue.py
caseylitton: client_info['uuid_cnt'] would return a number so we'd want, false or client_info['uuid_cnt'] > 0 or not client_info['results']

I agree this can be cleaned up.  Probably with an any() and a better boolean expression like you suggested.  I'll keep this in mind for the follow up PR given the various queue types.

src/snovault/elasticsearch/simple_queue.py
caseylitton: Haha, you like this these in-line conditionals...When I'm in a hurry I'll just set the var in long form conditionals and return at the end.  Yea, could be cleaned up for sure.

PR #152, opened 2018-11-05
SNO-66 Add new endpoint to map schema @types to schema titles

forresttanaka:
Add a new endpoint, `/profiles-titles/`, that returns an object keyed by every @type in the system, and the corresponding `title` property from the schema, if available. This snovault branch is required for the ENCD-4033-search-header-profile branch.

### DIFF
--- a/src/snovault/schema_views.py
+++ b/src/snovault/schema_views.py
@@ -13,6 +13,7 @@ def includeme(config):
     config.add_route('schema', '/profiles/{type_name}.json')
     config.add_route('schemap', '/profiles/{type_name}')
     config.add_route('schemas_map', '/profiles-map/')
+    config.add_route('schemas_titles', '/profiles-titles/')
     config.scan(__name__)
 
 
@@ -81,3 +82,16 @@ def schemas_map(context, request):
             profiles_map[type_info.name] = type_info.schema['id']
     profiles_map['@type'] = ['JSONSchemas']
     return profiles_map
+
+
+@view_config(route_name='schemas_titles', request_method='GET',
+             decorator=etag_app_version_effective_principals)
+def schemas_map(context, request):
+    """Return mapping of all schema @types and their corresponding titles"""
+    types = request.registry[TYPES]
+    profiles_titles = {}
+    for type_info in types.by_item_type.values():
+        if 'title' in type_info.schema:
+            profiles_titles[type_info.name] = type_info.schema['title']
+    profiles_titles['@type'] = ['JSONSchemas']
+    return profiles_titles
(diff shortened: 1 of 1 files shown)


src/snovault/schema_views.py
```diff
@@ -81,3 +82,16 @@ def schemas_map(context, request):
...
     profiles_map['@type'] = ['JSONSchemas']
     return profiles_map
+
+
+@view_config(route_name='schemas_titles', request_method='GET',
+             decorator=etag_app_version_effective_principals)
+def schemas_titles(context, request): # pylint: disable=unused-argument
+    '''Return mapping of all schema @types and their corresponding titles'''
+    types = request.registry[TYPES]
+    profiles_titles = {}
```
caseylitton: This could be dict comprehension
```
profiles_titles = {
    type_info.name, type_info.schema['title']
    for type_info in types.by_item_type.values()
    if 'title' in type_info.schema
}
```

src/snovault/schema_views.py
```diff
@@ -81,3 +82,16 @@ def schemas_map(context, request):
             profiles_map[type_info.name] = type_info.schema['id']
     profiles_map['@type'] = ['JSONSchemas']
     return profiles_map
+
+
+@view_config(route_name='schemas_titles', request_method='GET',
+             decorator=etag_app_version_effective_principals)
+def schemas_titles(context, request): # pylint: disable=unused-argument
```
caseylitton: Inline comments technically get two or more spaces before the hash.
https://www.python.org/dev/peps/pep-0008/#inline-comments

Not a big deal, probably wouldn't have mentioned it if there wasn't another update.

forresttanaka: Thanks for the citation — I looked specifically for PEP8’s opinion on that and somehow never found that.

caseylitton: yea, I wouldn't have known except some version of text editor I had in the past flagged it.

src/snovault/schema_views.py
```diff
@@ -81,3 +82,17 @@ def schemas_map(context, request):
...
+             decorator=etag_app_version_effective_principals)
+def schemas_titles(context, request):  # pylint: disable=unused-argument
+    '''Return mapping of all schema @types and their corresponding titles'''
+    types = request.registry[TYPES]
+    profiles_titles = {
+        type_info.name: type_info.schema['title']
+        for type_info in types.by_item_type.values()
+        if 'title' in type_info.schema
+    }
+    profiles_titles['@type'] = ['JSONSchemas']
```
zoldello: Seems weird to have a set with only one object. Albeit I have seen similar code in other places prior.

PR #156, opened 2018-11-27
SNO-69-refactor-result-from-set_filters

zoldello:


src/snovault/helpers/helper.py
```diff
@@ -268,18 +268,16 @@ def build_terms_filter(query_filters, field, terms):
...
     query_filters = query['post_filter']['bool']
     used_filters = {}
-    if static_items is None:
-        static_items = []
+    result_filters = []
 
     # Get query string items plus any static items, then extract all the fields
     qs_items = list(request.params.items())
-    total_items = qs_items + static_items
+    total_items = qs_items + (static_items or [])
```
caseylitton: I'm not a fan of the ```+``` for list create/concatenation unless it is being used a lot in the function.  Here it's just a shortcut to create a new list from two other lists. 

I'd prefer to make it more readable with explicit steps.
```
total_items = qs_items.copy()
if static_items:
    total_items.extend(static_items)
```
    
The original code was fine too, so may be best to not change in any case.

src/snovault/helpers/helper.py
```diff
@@ -268,18 +268,16 @@ def build_terms_filter(query_filters, field, terms):
         query_filters['must'].append(filter_condition)
 
 
-def set_filters(request, query, result, static_items=None, filter_exclusion=None):
-    """
-    Sets filters in the query
-    """
+def set_filters(request, query, static_items=None, filter_exclusion=None):
```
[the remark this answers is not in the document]
zoldello: How important is having query_filters passed in a an argument to you? The method- set_filters is used outside of snovalt and already has four arguments; so adding query_filters as a parameters seems to give very little gain given what it takes to add it.

src/snovault/helpers/helper.py
zoldello: Agreed. Your way seems more pythonic. However, the code works and we are not trying to polish up set_filters. So it is best to leave it.

caseylitton: Hey, We need to return this to it's original form.  It was an unnecessary change and we're being more strict on that now.

zoldello:
You mentioned this: 
"Looks good. A little further clean up and maybe revert the unneeded change to the list concatenation."

Was that addressed. I took out the doc string comments.

.travis.yml
```diff
@@ -101,7 +101,8 @@ script:
       --browser-arg-int build  "$TRAVIS_BUILD_NUMBER" \
       --browser-arg-int idleTimeout 300 \
       --browser-arg name "$TRAVIS_REPO_SLUG $TRAVIS_BRANCH $TRAVIS_COMMIT" \
-      --browser-arg browser "$BROWSER"
+      --browser-arg browser "$BROWSER" \
+      --browser-arg platform "Linux"
```
caseylitton: revert this change and rebase since this is in snovault now.

zoldello: I removed the two commit related to this and then rebased on master.

caseylitton:
closed.  code is way out of date.

PR #161, opened 2018-12-09
SNO-XX Travis update karen winnt

caseylitton:


caseylitton:
implemented in another ticket

PR #162, opened 2018-12-10
SNO-73 Add uuid queue module

caseylitton:
test with failed redis indexing process switching to simple queue

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,6 +3,7 @@ addopts =
     --pyargs snowflakes.tests
     --pyargs snovault.tests
     --pyargs snovault.elasticsearch.tests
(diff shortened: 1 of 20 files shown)


pytest.ini
```diff
@@ -3,6 +3,7 @@ addopts =
     --pyargs snowflakes.tests
     --pyargs snovault.tests
     --pyargs snovault.elasticsearch.tests
+    # --pyargs snovault.elasticsearch.uuid_queue.tests
```
caseylitton: Redis is not yet setup for testing, redis server needs to be running locally.  Leaving this commented out till fixed.

keenangraham (approved):
We might need some high-level overview docs for this?

PR #165, opened 2019-01-29
SNO-76-Refactor-out-matrix-and-result

zoldello:


zoldello:
This is not needed anymore.

PR #168, opened 2019-02-14
SNO-80-78 typeahead for facets

emro:
We need a new facet type in the schema for typeahead facets.

### DIFF
--- a/src/snovault/helpers/helper.py
+++ b/src/snovault/helpers/helper.py
@@ -340,7 +340,7 @@ def build_aggregation(facet_name, facet_options, min_doc_count=0):
     agg_name = facet_name.replace('.', '-')
 
     facet_type = facet_options.get('type', 'terms')
-    if facet_type == 'terms':
+    if facet_type in ['terms', 'typeahead']:
         agg = {
             'terms': {
                 'field': field,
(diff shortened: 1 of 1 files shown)


src/snovault/helpers/helper.py
```diff
@@ -340,12 +340,16 @@ def build_aggregation(facet_name, facet_options, min_doc_count=0):
     agg_name = facet_name.replace('.', '-')
 
     facet_type = facet_options.get('type', 'terms')
-    if facet_type == 'terms':
+    if 'length' in facet_options:
+        facet_length = 3000 if facet_options['length'] == 'long' else 200
+    else:
+        facet_length = 200
+    if facet_type in ['terms', 'typeahead']:
```
forresttanaka: The one suggestion just if you feel like is to move the assignment on line 342 to just above 347, just so the assignment is just above where you check it. Technically, it really doesn’t matter though.

caseylitton:
Closing along with parent ticket in encoded, sno-78 probably being split out

emro:
I renamed the SNO-78 commit to remove quotes around "long" and I merged commits so we would have two, one for the SNO-80 change and one for the SNO-78 change.

PR #171, opened 2019-03-16
Make nginx dev config respect forwarded proto.

lrowe:
This allows proxying from another nginx server useful when wanting to expose current development intstance through a cloud instance.

### DIFF
--- a/src/snovault/nginx-dev.conf
+++ b/src/snovault/nginx-dev.conf
@@ -11,6 +11,12 @@ http {
         server 127.0.0.1:6543;
         keepalive 10;
     }
+
+    map $http_x_forwarded_proto $forwarded_proto {
+        default $http_x_forwarded_proto;
+        ''      $scheme;
+    }
+
     server {
         listen  8000;
         location / {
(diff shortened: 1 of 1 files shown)


src/snovault/nginx-dev.conf
```diff
@@ -11,6 +11,12 @@ http {
         server 127.0.0.1:6543;
         keepalive 10;
     }
+
+    map $http_x_forwarded_proto $forwarded_proto {
```
hitz: if I follow this it is just saying to use http/https based on value of http_x_forwarded_proto or scheme.  Although I am not sure how these are set.

lrowe: Yes. It's using the X-Forwarded-Proto header if it exists and the scheme if it doesn't. That header is set by a downstream proxy server (in my case the Google Cloud Integrated Authentication Proxy I use for authentication.) I needed this so I could test my authentication setup locally.

hitz:
@lrowe @stuartmiyasato @caseylitton @cricketsloan I think we have pulled nginx.conf out of everything and besides we don't use nginx->snovault.  So I think this should be rejected?  Convince me otherwise.

lrowe:
This `nginx-dev.conf` is only used in development when running `bin/dev-servers`. The change is only really applicable when you're developing authentication stuff, in my case I had the Google Cloud Auth proxy in front of cloud instance and wanted to proxy back to my local development machine over an ssh tunnel.

PR #172, opened 2019-03-16
Update versions, Elasticsearch 6, Python 3.7

lrowe:
Draft PR for visibility. This needs more testing from me before merging. Given the Elasticsearch version update it will require understanding the impact on your current infrastructure.


src/snowflakes/renderers.py
```diff
@@ -226,13 +227,13 @@ def should_transform(request, response):
         if request.authorization is not None:
             format = 'json'
         else:
-            mime_type = request.accept.best_match(
+            acceptable = request.accept.acceptable_offers(
```
hitz: This is a pyramid 1.10 update?

lrowe: WebOb update (Pyramid's request/response object.) Silencing a deprecation warning on the best_match method. https://docs.pylonsproject.org/projects/webob/en/stable/api/webob.html#webob.acceptparse.AcceptValidHeader.best_match

src/snovault/tests/elasticsearch_fixture.py
```diff
@@ -45,6 +46,15 @@ def server_process(datadir, host='127.0.0.1', port=9201, prefix='', echo=False):
...
 
+    es = Elasticsearch('%s:%s' % (host, port))
+    es.indices.put_template(name="default", order=0, body={
+        "index_patterns": ["*"],
+        "settings": {
+            "number_of_shards": "1",
+            "number_of_replicas": "0",
+        },
+    })
+
```
lrowe: This template doesn't do anything because they end up overridden by the per index settings in create_mapping. Doesn't seem to make much difference when running locally. (I don't think number_of_replicas has any effect in a single node cluster.) Not sure what to do to make tests run in Travis at this point (they seem to just get stuck.) May be worth exploring other options.

hitz:
@lrowe do you have a version of this that is ONLY Elasticsearch changes?  Is it for 6.x or 7.x?

lrowe:
@hitz not currently, though it looks like it should be easy to pick out the relevant commits. I'm currently running ES 6.6. Not sure what would be involved in updating to ES 7 though imagine this would get much of the way there. I am less confident of the region search ES changes than the other stuff as we're not yet using it. I will try and pull out the relevant commits into a current branch next week.

keenangraham:
Our search code has changed a lot from this PR. Namely we use `elasticsearch-py-dsl` to build up ES queries now, which is tied to an ES version. (We'll just need to bump both versions at the same time.)

The most important part from this PR is probably the `_all` deprecation. My only thought is that if we are going to use `sub_mapping['copy_to'] = 'full_text'` to replace it why not just call `full_text` -> `_all`. (This would prevent having to change code where this field is actually used.)

A lot of the other stuff looks like things we already merged in Pyramid/Python upgrade.

PR #178, opened 2019-04-08
SNO-86-escape-search-text

zoldello:


zoldello:
@hitz Yes, I tested and retested it after seeing your comment. The ENCODED app fails when there are two or more question marks in a search string. This reduces the number of the question marks to one. The solution is fine to me but it would be nice to get a second opinion. Is this a good enough solution or should this part of the ticket be held off until I get to the root cause of the failure?

.travis.yml
```diff
@@ -9,7 +9,7 @@ cache:
 addons:
   apt:
     packages:
-    - oracle-java8-set-default
+    - oracle-java9-set-default
```
caseylitton: revert before moving to merge ready

zoldello: @caseylitton Done. Thanks for the catch.

zoldello:
@caseylitton I do not see my comment to you earlier. I removed the java reference.

PR #179, opened 2019-04-08
SNO-87-fixed-issue-while-profile-url-does-not-work

zoldello:


caseylitton:
fixed-issue-while-profile-url-does-not-work   

Don't write in the past tense.

PR #183, opened 2019-04-23
SNO-90-added-advancedQuery

zoldello:


zoldello:
@hitz This code was modified a bit from the last review.

PR #191, opened 2019-05-13
SNO-93-user-count-does-not-match

zoldello:


src/snovault/viewconfigs/report.py
```diff
@@ -52,7 +52,7 @@ def preprocess_view(self, views=None, search_result_actions=None):
         res = SearchView(self._context, self._request).preprocess_view(
             views=views,
             search_result_actions=search_result_actions,
-            preserve_order=True,
+            preserve_order=False,
```
hitz: I think this should still be left as True, unless you have some reason why it's causing a problem.

PR #192, opened 2019-05-14
SNO-144-redirect-downloads-through-nginx-or-apache

lrowe:
Instead of hard coding X-Accel-Redirect make it configurable so Apache mod_wsgi Location header can be used.
When unset proxy through Python process.

### DIFF
--- a/src/snovault/attachment.py
+++ b/src/snovault/attachment.py
@@ -37,6 +37,13 @@ def includeme(config):
     config.scan(__name__)
 
 
+class InternalRedirect(Response):
+    @staticmethod
+    def _make_location_absolute(environ, value):
+        # Avoid changing Location header so Apache internally redirects.
+        return value
+
+
 def parse_data_uri(uri):
     if not uri.startswith('data:'):
         raise ValueError(uri)
@@ -240,9 +247,10 @@ def download(context, request):
 
     # If blob is external, serve via proxy using X-Accel-Redirect
     blob_storage = request.registry[BLOBS]
-    if hasattr(blob_storage, 'get_blob_url'):
+    accel_redirect_header = request.registry.settings.get('accel_redirect_header')
+    if hasattr(blob_storage, 'get_blob_url') and accel_redirect_header:
         blob_url = blob_storage.get_blob_url(download_meta)
-        return Response(headers={'X-Accel-Redirect': '/_proxy/' + str(blob_url)})
+        return InternalRedirect(headers={accel_redirect_header: '/_proxy/' + str(blob_url)})
 
     # Otherwise serve the blob data ourselves
     blob = request.registry[BLOBS].get_blob(download_meta)
--- a/base.ini
+++ b/base.ini
@@ -8,6 +8,7 @@ elasticsearch.server = localhost:9201
 aws_ip_ranges_path = %(here)s/aws-ip-ranges.json
 download_proxy = https://download.encodeproject.org/
 annotations_path = %(here)s/annotations.json
+accel_redirect_header = X-Accel-Redirect
 
 # Only run ec2metadata on ec2 instances
 # XXX really need to reorganise ini files for more reuse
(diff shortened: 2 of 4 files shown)


lrowe:
Note: will require adding the setting to ini files in encoded.

hitz:
@lrowe This looks pretty straightforward.  Is there anything to test?   I assume @cricketsloan still wants it?

lrowe:
I think this is generally useful as it makes it possible to host only in Apache. Do make sure that the configuration change makes it through to encoded too as without it the file is proxied through Python.

To test curl a localhost download url directly on the server. You should see an empty response with a an X-Accel-Redirect header. Test when accessing through Nginx that the file is downloaded.

keenangraham:
This looks okay but trying to trace through exactly what's happening. I'm assuming the `_make_location_absolute` is overriding this Webob response method: https://github.com/Pylons/webob/blob/6798afbce84d251b6aa1ba228acb150a2f03e143/src/webob/response.py#L1313

This looks like it is called from `_abs_headerlist`: https://github.com/Pylons/webob/blob/6798afbce84d251b6aa1ba228acb150a2f03e143/src/webob/response.py#L1325 if `location` is in a header key.

When I add a print statement and download a document from a demo I don't actually see this method getting triggered despite returning the `InternalRedirect` response:

```
def _make_location_absolute(environ, value):
        # Avoid changing Location header so Apache internally redirects.
        print('in _make_location_absolute', value)
        return value
```

I suppose the step I'm missing is that _instead_ of passing `X-Accel-Redirect` (which is an NGINX directive) you would pass `Location` which would trigger the Apache rewrite rule on /proxy? But you have to prevent the absolute URI from getting prepended so it stays internal? 

So to actually get this to do anything you would have to set `accel_redirect_header = X-Accel-Redirect` to `accel_redirect_header = Location`? In which case why call it accel_redirect?

What am I missing @lrowe?

lrowe:
> So to actually get this to do anything you would have to set accel_redirect_header = X-Accel-Redirect to accel_redirect_header = Location?

Yes. I currently deploy encoded without nginx in front so would like the option to have the internal redirect happen in Apache.

As a bonus development instances can now run without nginx and will follow the redirect in Python.

> In which case why call it accel_redirect?

I couldn't think of a better name for what this does. The Location header behaviour is buried deep in the CGI spec. Perhaps `internal_redirect_header`?

keenangraham:
By the way did you do something similar for files on your system? https://github.com/ENCODE-DCC/encoded/blob/ecfb135ba84ef5cad2a71638720d782cbfc4d14a/src/encoded/types/file.py#L789

lrowe:
@keenangraham just made encode-dcc/encoded#3182 with the encoded change.

keenangraham:
Rebased.

src/snovault/attachment.py
```diff
@@ -37,6 +37,13 @@ def includeme(config):
     config.scan(__name__)
 
 
+class InternalRedirect(Response):
+    @staticmethod
+    def _make_location_absolute(environ, value):
+        # Avoid changing Location header so Apache internally redirects.
```
caseylitton: It looks like we're overriding the _make_location_absolute for downloads with 'get_blob_url' and 'accel_redirect_header'.  

The override is skipping this line of code in webob, https://github.com/Pylons/webob/blob/master/src/webob/response.py#L1314

What is the difference between returning the value regardless of the condition, 'if SCHEME_RE.searchv(value)' ?

lrowe:
```python
    @staticmethod
    def _make_location_absolute(environ, value):
        if SCHEME_RE.search(value):
            return value

        new_location = urlparse.urljoin(_request_uri(environ), value)
        return new_location
```
Looks like the `if SCHEME_RE.searchv(value)` is just a fast path for values which are already absolute urls (so would not be changed by urljoin.)

PR #194, opened 2019-05-16
SNO-92-fix-batch-upgrade-logging

caseylitton:


caseylitton:
going in with sno100 
https://github.com/ENCODE-DCC/snovault/pull/201

PR #197, opened 2019-05-21
SNO-97-matrix-update

zoldello:


src/snovault/helpers/helper.py
```diff
@@ -281,6 +281,7 @@ def set_filters(request, query, result, static_items=None, filter_exclusion=None
 
     # Get query string items plus any static items, then extract all the fields
     qs_items = list(request.params.items())
+    qs_items.extend(implicit_facet_terms or [])
```
caseylitton: if implicit_facet_terms then extend.  We don't need to extend with an empty list.

zoldello: @caseylitton I have this as a PR because I needed it for building a demo to prototype. I can add ed "do not merge" tag or close it if you like.

qs_items.extend(implicit_facet_terms or []), I did it this way because if implicit_facet_terms == None, extend-method crashes. Using the "or []"  makes it do nothing without crashing. I can add an if-statement if that makes it easier to read.

zoldello: @caseylitton I thought about it. I added an if-statement. As-is, it is a bit confusing why there is the "or []".

I also will close this PR. It is not ready for release.

zoldello:
Not ready for release/code-review.

PR #202, opened 2019-06-13
SNO-99 port travis tests to circleci

paul-sud:


p-sud:
@caseylitton There are two commits in here, the first is a test config using oracle jdk 11, the second changes the config to use openjdk

.circleci/config.yml
```diff
@@ -0,0 +1,91 @@
...
+              bsdtar \
+              elasticsearch \
+              graphviz \
+              nodejs \
+              openjdk-11-jdk \
+              postgresql-9.3 \
+              ruby2.3 \
+              ruby2.3-dev
+            sudo chown -R circleci /etc/elasticsearch
+            sed -i '1s;^;export PATH=/usr/share/elasticsearch/bin:/usr/lib/postgresql/9.3/bin:$PATH\n;' $BASH_ENV
```
caseylitton: we're updating to postgres 11 at some point in the near future.  It has a different installation and some of the directories change.  Is it possible to make some of these strings into variables?
```
PATH=%(ES_BIN)s:%(PG_BIN)s:$PATH
```
or maybe for the node url above
```
echo "%(NODE_URL)s stable main"
```
Just curious how we may accomplish this.  We have similar lines in all the cloud configs so it might be nice to have one version var file...

p-sud: It's definitely possible, the first method that comes to mind is environment variables. (see https://circleci.com/docs/2.0/env-vars/ ). You could read them in from a file for sure as well

caseylitton: cool.  We're not ready for this but lets try to keep it in mind. It would be nice to eventually integrate this with cloud configs.

caseylitton: nvm, you already did it. :)

p-sud: Saw your comment too late haha. ES, Postgres, and Node version numbers are now environment variables and not hard coded, so they'll be easier to change later

caseylitton (approved):

PR #208, opened 2019-07-05
SNO-104-disable-boost-values

zoldello:


hitz:
I think you have to set include_in_all = True as default somewhere?  Or is this the ES default?

zoldello:
> I think you have to set include_in_all = True as default somewhere? Or is this the ES default?

include_in_all is the default as of ES5. I can check if it is true for ES6 and higher

PR #209, opened 2019-07-15
SNO-106-fix-spec-char-500-error

zoldello:


src/snovault/helpers/helper.py
```diff
@@ -75,33 +75,29 @@ def prepare_search_term(request):
...
-            return '*'
-        # elasticsearch uses ':' as field delimiter, but we use it as namespace designator
-        # if you need to search fields you have to use @type:field
-        # if you need to search fields where the field contains ":", you will have to escape it
-        # yourself
-        search_term = search_term.replace('/', r'\/').replace('^', r'').replace('<', r'').replace('>', r'')
-        search_term = re.sub('\?+', '?', search_term)
-        if search_term.find("@type") < 0:
-            search_term = search_term.replace(':', r'\:')
+    # elasticsearch uses ':' as field delimiter, but we use it as namespace designator
```
keenangraham: Let's get rid of this comment.

src/snovault/helpers/helper.py
```diff
@@ -75,33 +75,29 @@ def prepare_search_term(request):
...
-        try:
-            query = prefixfields('embedded.', advanced_query, dialects.elasticsearch)
-        except IllegalStateException:
-            msg = "Invalid query: {}".format(advanced_query)
-            raise HTTPBadRequest(explanation=msg)
-        else:
-            search_query = query.getText()
-    if search_term:
-        search_query = ''.join([search_query, ' And ', search_term]) if search_query else search_term
+    if advanced_query and search_term:
```
keenangraham: Actually this whole block can just be:

```
if advanced_query and search_term:
    search_query = ''.join([advanced_query, ' AND ', search_term])
else:
    search_query = advanced_query or search_term
```

We should use capital `AND` to be consistent with the operators here: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html

zoldello: fixed

src/snovault/helpers/helper.py
zoldello: Removed the comment.

PR #210, opened 2019-07-16
SNO-108-improve-search

keenangraham:
* Add word-delimiter and english stemmer to ES analyzer (fixes RNA-seq/rnaseq problem, and year/years problem).
* Reinstate query-time boost values so exact matches rank higher in results.
* Add file and lab to default doc types

### DIFF
--- a/src/snovault/helpers/helper.py
+++ b/src/snovault/helpers/helper.py
@@ -173,7 +173,7 @@ def get_search_fields(request, doc_types):
     for doc_type in doc_types:
         type_info = types[doc_type]
         for value, boost in type_info.schema.get('boost_values', {}).items():
-            fields.add('embedded.' + value + '^' + boost)
+            fields.add('embedded.' + value + '^' + str(boost))
             highlights['embedded.' + value] = {}
     return list(fields), highlights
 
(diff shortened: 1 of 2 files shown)


src/snovault/helpers/helper.py
```diff
@@ -172,8 +172,8 @@ def get_search_fields(request, doc_types):
     types = request.registry[TYPES]
     for doc_type in doc_types:
         type_info = types[doc_type]
-        for value in type_info.schema.get('boost_values', ()):
-            fields.add('embedded.' + value)
+        for value, boost in type_info.schema.get('boost_values', {}).items():
+            fields.add('embedded.' + value + '^' + str(boost))
```
hitz: so this theoretically makes the boost values > 1.0 actually do something?  I thought we were not going to do index-time boosting anyway...   Still it seems harmless.

keenangraham: This is query-time boosting. The fields get added to the query_string query like this example from ES docs: 
```
curl -X GET "localhost:9200/_search" -H 'Content-Type: application/json' -d'
{
    "query": {
        "query_string" : {
            "fields" : ["content", "name^5"],
            "query" : "this AND that OR thus",
        }
    }
}
```
Since mostly all of these fields are keyword fields the boost comes into play when there is an exact match. This helps when searching for accession for example. The actual item with that accession gets ranked higher. Also when searching for biosample age, which otherwise can get lost in results that match on more fields but have the wrong age.

keenangraham:
Closing for SNO-108-improve-search-new

PR #216, opened 2019-09-06
SNO-112-fix-npm-audit-issues

zoldello:


hitz:
@forresttanaka @caseylitton  I think it's fine to have a massive update of package-lock.json, since it only effects "snowflakes", which is non-production.  I'll have reviewer remember to test a local version.

gulpfile.js
```diff
@@ -1,34 +1,38 @@
-var gulp = require('gulp');
-var gutil = require('gulp-util');
-var webpack = require('webpack');
+const gulp = require('gulp');
+const log = require('fancy-log');
```
emro: Just curious, why did you replace `gutil` with `log` here but not in Encode?

gulpfile.js
```diff
@@ -1,34 +1,38 @@
...
+const setProduction = function (cb) {
   process.env.NODE_ENV = 'production';
-});
+  if (cb) {
+      cb();
+  }
+};
 
-var webpackOnBuild = function (done) {
+const webpackOnBuild = function (done) {
```
emro: This could be written as an arrow function, `done => (err, stats) => {` etc

.babelrc
```diff
@@ -1,4 +1,11 @@
 {
-	presets: ['es2015', 'react'],
-	plugins: ['transform-object-rest-spread']
-}
+	presets: [
+		'@babel/preset-env',
+		'@babel/preset-react',
+		'@babel/flow',
```
emro: It looks like `@babel/flow` helps with type checking? Would this be helpful on Encode as well?

zoldello: Maybe. We can discuss it.

gulpfile.js
zoldello: I was hesitant to do that because of functional context (this-keyword)- It is handled different between function and arrow function. I did not want to have to worry about that along with the many other things I had to worry about.

emro: OK! I was comparing your two npm package updates and noticed that you made this change in the other one: https://github.com/ENCODE-DCC/encoded/pull/2913/commits/9ac80faf5c5979a3bca21eb6cb3b8f80e113343d#diff-b9e12334e9eafd8341a6107dd98510c9L14

gulpfile.js
zoldello: I do not remember the detail. gutil is deprecated (https://medium.com/gulpjs/gulp-util-ca3b1f9f9ac5). So eventually should be replaced. I can replace it in ENCODE if you desire.

emro: It might be a good time to replace it in Encode if we'll have to replace it eventually.

zoldello: Replaced.

gulpfile.js
zoldello: Since the code works and the change is not significant, I left it as-is. It can be updated in a future release.

PR #217, opened 2019-09-06
SNO-113-make-audit-view

keenangraham:
Adds auditv2 backend.


src/snovault/elasticsearch/searches/queries.py
```diff
@@ -1000,3 +1001,20 @@ def build_query(self):
...
         return self.search
+
+
+class AuditMatrixQueryFactoryWithFacets(BasicMatrixQueryFactoryWithFacets):
+    '''
+    Like BasicMatrixQueryFactoryWithFacets but adds aggregation for all audit fields.
+    '''
+
+    def _get_matrix_definition_name(self):
+        return self.kwargs.get('matrix_definition_name') or AUDIT
```
hitz: in what cases will matrix_definition_name get set?

src/snowflakes/search_views.py
```diff
@@ -36,7 +38,8 @@ def includeme(config):
     config.add_route('report', '/report{slash:/?}')
     config.add_route('matrixv2_raw', '/matrixv2_raw{slash:/?}')
```
hitz: Do we still need this?

src/snowflakes/tests/test_searchv2.py
```diff
@@ -465,3 +465,54 @@ def test_collection_listing_db_view(workbook, testapp):
     assert r.json['@type'] == ['SnowflakeCollection', 'Collection']
     assert r.json['@context'] == '/terms/'
     assert r.json['description'] == 'Listing of Snowflakes'
+
+
+def test_auditv2_response(workbook, testapp):
```
hitz: feel like the v2s should be removed from test function names.

src/snowflakes/tests/test_searchv2.py
```diff
@@ -465,3 +465,54 @@ def test_collection_listing_db_view(workbook, testapp):
...
+    assert r.json['matrix']['x']['group_by'] == 'snowflakes.type'
+    assert r.json['matrix']['audit.ERROR.category']
+    assert r.json['matrix']['audit.WARNING.category']
+    assert r.json['matrix']['audit.NOT_COMPLIANT.category']
+    assert r.json['matrix']['audit.INTERNAL_ACTION.category']
+    assert 'search_base' in r.json
+    assert r.json['search_base'] == '/search/?type=Snowball'
+
+
+def test_auditv2_response_with_search_term_type_only_clear_filters(workbook, testapp):
```
hitz: v2

src/snowflakes/tests/test_searchv2.py
```diff
@@ -465,3 +465,54 @@ def test_collection_listing_db_view(workbook, testapp):
...
+    r = testapp.get('/audit/?type=Snowball&searchTerm=crisp')
+    assert 'clear_filters' in r.json
+    assert r.json['clear_filters'] == '/audit/?type=Snowball'
+
+
+def test_auditv2_response_debug(workbook, testapp):
```
hitz: v2

src/snowflakes/tests/test_searchv2.py
```diff
@@ -465,3 +465,54 @@ def test_collection_listing_db_view(workbook, testapp):
...
+    r = testapp.get('/audit/?type=Snowball&debug=true')
+    assert 'debug' in r.json
+
+
+def test_auditv2_response_no_results(workbook, testapp):
```
hitz: v2

src/snowflakes/tests/test_searchv2.py
```diff
@@ -465,3 +465,54 @@ def test_collection_listing_db_view(workbook, testapp):
...
+    r = testapp.get(
+        '/audit/?type=Snowball&status=no_status',
+        status=404
+    )
+    assert r.json['notification'] == 'No results found'
+
+
+def test_auditv2_view_no_matrix_defined(workbook, testapp):
```
hitz: v2

hitz (approved):
Code here is generally good.   I think that the tests functions with v2 in the name should probably be changed.   I suspect that we will have to fix at least 1 bug in downstream encoded tickets, so this should probably be update?

src/snovault/elasticsearch/searches/queries.py
keenangraham: For example the summary view is just a matrix view with the definition set to summary: https://github.com/ENCODE-DCC/snovault/blob/32a140e46036be806321f837ba0e3cf570e918f7/src/snowflakes/search_views.py#L204

This allows you to add multiple definitions to the item type under different names that use the same underlying code.

So all you would need to do to make, say, an audit by lab (instead of assay title) view is create a new definition on the type and pass it in to the view. You get to reuse all of the backend code.

src/snowflakes/search_views.py
keenangraham: Yes, most importantly these are demo implementations for others to learn from (and to remind myself in the future). Some of them might be useful.

src/snowflakes/tests/test_searchv2.py
keenangraham: I've left the v2s to be consistent with the file name. This is something we can unwind when we delete old search code next round.

PR #238, opened 2019-10-26
SNO-125-switch-batch-download-to-query-string

keenangraham:


src/snovault/elasticsearch/searches/parsers.py
```diff
@@ -323,3 +323,109 @@ def split_filters_by_must_and_exists(self, params=None):
...
+        Copies request and fills in new deduplicated query_string.
+        '''
+        if deduplicate:
+            self.deduplicate()
+        request = self._request.copy()
+        request.query_string = self.get_query_string()
+        return request
+
+
+class QueryString(MutableParamsParser):
```
hitz: I am wondering if this is generalizable outside of ES queries?   Probably not.

keenangraham: Yes! I wrote this exactly to be a general query string parser/manipulator. For example Phil's use case of adding assay_title to the request before passing it to search endpoint in theoretical hardcoded target_matrix endpoint.

PR #242, opened 2019-11-20
SNO-126-allow-default-value-for-missing-aggregation-field

keenangraham:


hitz:
Note: branch not deleted because ENCD tickets depend on it until snovault release.

PR #245, opened 2019-12-14
SNO-128-make-select-calculated-properties-view

keenangraham:


keenangraham:
Looks like tests need same version fix as encoded.

PR #249, opened 2020-01-23
SNO-132 Upgrade python related deps

caseylitton:
Look at commits, there are a couple different updates

This goes with encd pr

ENCD-3983 Update to ubuntu 18 / SNO-132 Update to python3.6 #3109
https://github.com/ENCODE-DCC/encoded/pull/3109


base.ini
```diff
@@ -1,7 +1,7 @@
 [app:app]
 use = egg:snovault#snowflakes
 sqlalchemy.url = postgresql:///snowflakes
-tm.attempts = 3
+retry.attempts = 3
```
hitz: what is this parameter for?

src/snovault/nginx-dev.conf
```diff
@@ -11,6 +11,12 @@ http {
         server 127.0.0.1:6543;
```
hitz: This should probably be removed from snovault.

src/snovault/tests/elasticsearch_fixture.py
```diff
@@ -19,7 +19,7 @@ def server_process(datadir, host='127.0.0.1', port=9201, prefix='', echo=False):
         print('IN TRAVIS')
```
hitz: Travis is not used, no longer relevant claus

src/snovault/tests/elasticsearch_fixture.py
```diff
@@ -19,7 +19,7 @@ def server_process(datadir, host='127.0.0.1', port=9201, prefix='', echo=False):
...
         echo = True
         args.append('-Epath.conf=%s/conf' % os.environ['TRAVIS_BUILD_DIR'])
-    else:
+    elif os.path.exists('/etc/elasticsearch'):
         print('NOT IN TRAVIS')
         args.append('-Epath.conf=./conf')
```
hitz: what happens if /etc/elasticsearch doesn't exist?

src/snovault/tests/test_snowflake_hash.py
```diff
@@ -10,4 +10,4 @@
 @pytest.mark.parametrize(('password', 'pwhash'), TEST_HASHES.items())
```
hitz: I feel like we don't need this at all any longer.

src/snovault/tests/testing_views.py
```diff
@@ -417,12 +417,12 @@ def testing_retry(context, request):
     from transaction.interfaces import TransientError
 
     model = context.model
-    request._attempt = getattr(request, '_attempt', 0) + 1
+    attempt = request.environ.get('retry.attempt')
```
hitz: oh this must  be the replacement for tm.attempts.

src/snowflakes/tests/test_access_key.py
```diff
@@ -137,4 +137,4 @@ def test_access_key_uses_snowflake_hash(collections, access_key):
     from snovault.snowflake_hash import SNOWHash
```
hitz: again I don't think this hash encryption has been used for years

src/snowflakes/types/access_key.py
```diff
@@ -87,14 +86,14 @@ def access_key_add(context, request):
...
-            for principal in effective_principals(request)
+            for principal in request.effective_principals
             if principal.startswith('userid.')
         ]
 
     password = None
     if 'secret_access_key_hash' not in request.validated:
         password = generate_password()
-        request.validated['secret_access_key_hash'] = crypt_context.encrypt(password)
+        request.validated['secret_access_key_hash'] = crypt_context.hash(password)
```
hitz: this one IS used though

src/snovault/tests/testing_views.py
caseylitton: this is spelled wrong, 

```'retry.attempt'```
vs
```retry.attempts = 3```
in the config.  Which is an update in encoded too.

ya know what's weird, the region indexer issue says it's failing do to max retrys.  Wonder if this is related.

caseylitton: Well, it only shows up in test code though.

base.ini
caseylitton: I've only seen it in tests, we have a test making sure it's 3.

PR #257, opened 2020-03-03
[TMP] live update to fix _stats

caseylitton:


hitz (changes requested):
just move stats definition out of the if (), I think this will be OK.

caseylitton:
way old

PR #264, opened 2020-03-26
SNO-143 switch update readmes

caseylitton:


README.md
```diff
@@ -1,10 +1,10 @@
 SnoVault JSON-LD Database Framework
 ===================================
 
-## System Installation (OSX Catlina 10.15.2)
-    The instructions below are not guaranteed to work.  See the dependency's website for
-    furthur instruction.  We will try to keep this updated as OSX, Xcode, and brew update.  These
-    instructions were also tested on Mojave 10.14.6.
+## System Installation (OSX Catlina~10.15.2, Mojave~10.14.6)
+    We will try to keep this updated as OSX, Xcode, and brew update.  Howere, The steps below are
```
hitz, line 5: However spelled wrong.

hitz (changes requested):

README.md
```diff
@@ -1,10 +1,10 @@
 SnoVault JSON-LD Database Framework
 ===================================
 
-## System Installation (OSX Catlina 10.15.2)
-    The instructions below are not guaranteed to work.  See the dependency's website for
-    furthur instruction.  We will try to keep this updated as OSX, Xcode, and brew update.  These
-    instructions were also tested on Mojave 10.14.6.
+## System Installation (OSX Catlina~10.15.2, Mojave~10.14.6)
```
zoldello, line 4: It is fine if you want to word it this way. But technically, these steps are guaranteed to work for the OS/packages version listed and not guaranteed for anything outside this scope.

src/snovault/elasticsearch/indexer.py
```diff
@@ -366,7 +366,7 @@ def check_log_indexing_times(self, update_infos):
             with open(self.indexer_initial_log_path, 'w', encoding='utf-8') as file_handler:
                 for update_info in update_infos:
                     str_update_info = json.dumps(update_info, ensure_ascii=False)
-                    file_handler.write(str_update_info)
+                    file_handler.write(str_update_info + '\n')
```
zoldello, line 369: Since we are on Python 3.6+, in the future, you can you f-string (https://www.geeksforgeeks.org/formatted-string-literals-f-strings-python/)

setup.py
```diff
@@ -3,7 +3,7 @@
 from setuptools import setup, find_packages
 
 here = os.path.abspath(os.path.dirname(__file__))
-README = open(os.path.join(here, 'README.rst')).read()
+README = open(os.path.join(here, 'README.md')).read()
```
zoldello, line 6: Any reason for going tom .rst to md. I thought rst is common with python projects.

zoldello (approved):
My comments are general observations and a question. They can be deemed irrelevant as you see fit.

README.md
caseylitton: The lag between Us updating our README to match a dependency will always be greater than the Dependency updating it's readme.

src/snovault/elasticsearch/indexer.py
caseylitton: oh yea, I forgot about this.  We should start moving toward using that in new code.

setup.py
caseylitton: encoded was md, so I made this follow suite.

hitz (approved):

PR #266, opened 2020-03-27
SNO-150-fix-antlr-version

keenangraham:


keenangraham:
> Commit msg should have been 'SNO-150-fix-antlr-version'. not sure if this still messes up the merge message.

Disagree and not going to follow this practice. Whoever squash merges should copy and paste the branch name in if it's not correct. We are removing helpful information for the reviewer otherwise.

PR #272, opened 2020-05-27
SNO-154-fix-magic-json-bug

caseylitton:


keenangraham (approved):
Would be nice to test good and bad JSON in attachment tests: https://github.com/ENCODE-DCC/snovault/blob/dev/src/snovault/tests/test_attachment.py

caseylitton:
> Would be nice to test good and bad JSON in attachment tests: https://github.com/ENCODE-DCC/snovault/blob/dev/src/snovault/tests/test_attachment.py

Yes, Good idea.

keenangraham (approved):

PR #275, opened 2020-06-17
SNO-157-make-facets-collapsable

zoldello:


src/snovault/elasticsearch/searches/interfaces.py
```diff
@@ -55,6 +55,7 @@
 NOT_FLAG = '!'
 NOTIFICATION = 'notification'
 OBJECT_FRAME = 'object'
+OPEN_ON_LOAD = 'openOnLoad'
```
forresttanaka: Should this be `open_on_load` instead of `openOnLoad`? Some companies’ JSON does use camel case (Google I believe) and some use snake. But I think this would be the first JSON key in encoded that uses camel case.

forresttanaka (changes requested):
I don’t know if you’ve talked about the camel case vs snake case with others and agreed on camel case. So let me know if you wanted to stick with camel case. The other thing, more on the encoded side, is I usually see boolean `true` and `false` instead of the string “True” in schemas, like the `uniqueItems` property. Should that be the case here too?

forresttanaka (approved):

src/snovault/elasticsearch/searches/mixins.py
```diff
@@ -109,7 +113,8 @@ def _format_aggregation(self, facet_name):
             TERMS: self._get_aggregation_bucket(facet_name),
             TOTAL: self._get_aggregation_total(facet_name),
             TYPE_KEY: self._get_facet_type(facet_name),
-            APPENDED: JS_FALSE
+            APPENDED: JS_FALSE,
+            OPEN_ON_LOAD: self._get_facet_open_on_load(facet_name),
```
keenangraham, line 117: Should also update tests for `_format_aggregation` and `_format_aggregations`: 
https://github.com/ENCODE-DCC/snovault/blob/84c439a24587354903fff5977025795198ce1112/src/snovault/tests/test_searches_mixins.py#L1559
https://github.com/ENCODE-DCC/snovault/blob/84c439a24587354903fff5977025795198ce1112/src/snovault/tests/test_searches_mixins.py#L1592

I think you can do this by adding a `open_on_load: true/false` to the facet fixture: https://github.com/ENCODE-DCC/snovault/blob/84c439a24587354903fff5977025795198ce1112/src/snovault/tests/test_searches_mixins.py#L5

And then asserting that expected['open_on_load'] == actual['open_on_load'] in the tests.

src/snovault/elasticsearch/searches/mixins.py
```diff
@@ -56,6 +57,9 @@ def _get_facet_title(self, facet_name):
     def _get_facet_type(self, facet_name):
         return self._get_facets().get(facet_name, {}).get(TYPE_KEY, TERMS)
 
+    def _get_facet_open_on_load(self, facet_name):
```
keenangraham, line 60: This looks okay but need to add test for `_get_facet_open_on_load` in `snovault/tests/test_searches_mixins.py`. 

Should be called`test_searches_mixins_aggs_to_facets_mixin_get_facet_open_on_load`.

Can see `test_searches_mixins_aggs_to_facets_mixin_get_facet_type` for example of how to mock response from `_get_facets`: https://github.com/ENCODE-DCC/snovault/blob/84c439a24587354903fff5977025795198ce1112/src/snovault/tests/test_searches_mixins.py#L1458

keenangraham (changes requested):

keenangraham (approved):
Looks good. Thanks for adding tests.

keenangraham (approved):

PR #282, opened 2020-07-09
SNO-159-allow-remote-indexing

caseylitton:


src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+        )
+    except:
+        remote_indexing_threshold = default_threshold
+    if uuid_count < default_threshold:
+        return False
+    return True
+
+
+def _get_sqs_arn(queue_name):
+    boto_session = boto3.Session(region_name = 'us-west-2')
```
hitz: Should region_name be hardcoded here?

src/snovault/elasticsearch/indexer_state.py
```diff
@@ -38,6 +46,100 @@ def includeme(config):
     config.add_route('_indexer_state', '/_indexer_state')
     config.scan(__name__)
 
+
+def _get_this_instance_name(instance_name=None):
+    ec2 = boto3.client('ec2', region_name='us-west-2')
```
hitz: should region be hard coded?

src/snovault/elasticsearch/indexer_state.py
```diff
@@ -596,6 +698,12 @@ def indexer_state_show(request):
         if msg is not None:
             return msg
 
+    # requesting reset on remote indexing
+    reset_remote = request.params.get("reset_remote")
+    if reset_remote is not None:
+        setup_indexing_nodes(request, state, reset=True)
```
hitz, line 712: sort of a strange violation of object oriented encapsulation here, no?    I guess _indexer_state could be called before Indexer?  But usually indexer will set up the nodes?   I guess IndexerState object has no reference to the actual Indexer object...   Oh, I see Indexer now instantiates IndexerState... so maybe we can just move setup_indexing_nodes to indexer.py?

src/snovault/elasticsearch/indexer.py
```diff
@@ -44,6 +49,11 @@
 log = logging.getLogger('snovault.elasticsearch.es_index_listener')
 MAX_CLAUSES_FOR_ES = 8192
 DEFAULT_QUEUE = 'Simple'
+_SQS = boto3.client('sqs', region_name='us-west-2')
```
hitz: should region_name be hard coded?

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+        remote_indexing_threshold = int(
+            request.registry.settings.get(
+                'remote_indexing_threshold',
+                default_threshold,
+            )
+        )
+    except:
+        remote_indexing_threshold = default_threshold
+    if uuid_count < default_threshold:
+        return False
```
hitz, line 180: oh so the machine will locally index all batches of < remote_indexing_threshold??

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+        return None
+
+
+def _send_sqs_msg(
+        cluster_name,
+        remote_indexing,
+        body,
+        delay_seconds=0,
+        save_event=False,
+    ):
```
hitz: this gives me a flake8 error:   "continuation line with same indent as next logical line"

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+    error_str = None
```
hitz: flake8 error: 'error_str is assigned to and never used

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
-        flush = True
-    else:
+    indexer_state = IndexerState(
+        request.registry[ELASTIC_SEARCH],
+        request.registry.settings['snovault.elasticsearch.index'],
+        followups=stage_for_followup,
+    )
+    # Check remote indexing and set head/indexing nodes
+    this_node = indexer_state.get_obj(INDEXING_NODE_INDEX)
+    other_node = indexer_state.get_obj(HEAD_NODE_INDEX)
```
hitz, line 353: HEAD_NODE is the front end?

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+    remote_indexing = asbool(request.registry.settings.get('remote_indexing', False))
+    did_timeout = False
+    if remote_indexing:
+        this_node, other_node, continue_on, did_timeout = _get_nodes(request, indexer_state)
```
hitz, line 357: didn't we just get this_node and other_node from indexer_state 5 lines up?

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+        if not did_timeout and not continue_on:
+            # Node state implies we do not need to run any code right now
+            return {'this_node': this_node, 'other_node': other_node}
+    elif this_node and other_node:
+        # if remote node states exists and remote_indexing false then this is the remote indexer
+        this_node, other_node, continue_on, did_timeout = _get_nodes(request, indexer_state)
+        if not did_timeout and not continue_on:
+            # Node state implies we do not need to run any code right now
+            return {'this_node': this_node, 'other_node': other_node}
+
```
hitz, line 367: what is the state where remote_indexing is False and either of this_node/other nodes are also False?  I,.e, the implicit "else:"  ...  just jumps to line 499?

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+    def _load_indexing(request, session, connection, indexer_state):
```
hitz, line 368: move this function somewhere else, not in the middle of view method

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+                    # http://www.postgresql.org/message-id/CAHGQGwEtJCeHUB6KzaiJ6ndvx6EFsidTGnuLwJ1itwVH0EJTOA@mail.gmail.com
+                    if snapshot_id is None and not request.json.get('recovery', False):
+                        snapshot_id = connection.execute('SELECT pg_export_snapshot();').scalar()
+        return_now = False
+        return result, invalidated, flush, first_txn, snapshot_id, restart, xmin, return_now
 
-    indexing_update_infos = []
-    if invalidated and not dry_run:
+
+    def _run_indexing(
```
hitz, line 478: move this function somewhere else, not in the middle of view method.   Ah, I see this was refactored from the old code into a method, it's not strictly new.

src/snovault/elasticsearch/indexer.py
```diff
@@ -285,41 +458,124 @@ def index(request):
...
             try:
-                es.indices.flush_synced(index=RESOURCES_INDEX)  # Faster recovery on ES restart
+                request.registry[ELASTIC_SEARCH].indices.flush_synced(index=RESOURCES_INDEX)  # Faster recovery on ES restart
             except ConflictError:
                 pass
+        return result, indexing_update_infos
 
+    # Load indexing
+    output_tuple = _load_indexing(request, session, connection, indexer_state)
+    result, invalidated, flush, first_txn, snapshot_id, restart, xmin, return_now = output_tuple
```
hitz, line 545: usually this would be combined on one line, but it would be really long so...

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+        indexer_state.log_reindex_init_state()
+        # OPTIONAL: restart support
+        if restart:  # Currently not bothering with restart!!!
+            xmin = -1
+            invalidated = []
+        # OPTIONAL: restart support
+
+        result = indexer_state.get_initial_state()  # get after checking priority!
+
+        if xmin == -1 or len(invalidated) == 0:
```
hitz, line 381: Is this the "don't index anything" or "index everything" state?

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+                xmin=xmin,
+                last_xmin=last_xmin,
+            )
 
-        flush = False
-        if last_xmin is None:
-            result['types'] = types = request.json.get('types', None)
-            invalidated = list(all_uuids(request.registry, types))
+        if len(invalidated) > SEARCH_MAX:  # Priority cycle already set up
             flush = True
```
hitz, line 408: so this should be full reindex

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
-                if first_txn is not None:
-                    result['first_txn_timestamp'] = first_txn.isoformat()
 
-            if invalidated and not dry_run:
-                # Exporting a snapshot mints a new xid, so only do so when required.
-                # Not yet possible to export a snapshot on a standby server:
-                # http://www.postgresql.org/message-id/CAHGQGwEtJCeHUB6KzaiJ6ndvx6EFsidTGnuLwJ1itwVH0EJTOA@mail.gmail.com
-                if snapshot_id is None and not recovery:
-                    snapshot_id = connection.execute('SELECT pg_export_snapshot();').scalar()
+                invalidated = set(invalidated)  # not empty if API index request occurred
```
hitz, line 421: oh I see, this isn't new code it's just rearranged a little, you moved it into this function

src/snovault/elasticsearch/indexer.py
```diff
@@ -285,41 +458,124 @@ def index(request):
...
+            xmin,
+        )
+        if this_node:
+            if this_node['node_index'] == HEAD_NODE_INDEX:
+                # Check remote indexing
+                uuids_count = len(invalidated)
+                do_remote_indexing = _determine_indexing_protocol(request, uuids_count)
+                if did_timeout:
+                    log.warning(f"Remote indexing index: did_timeout.  Run on head node.")
+                    head_node, indexing_node, time_now, did_fail, is_indexing_node  = setup_indexing_nodes(
```
hitz, line 569: multiple flake8 issues, here, none of the variables are ever used and there are 2 spaces before =

src/snovault/elasticsearch/indexer.py
```diff
@@ -285,41 +458,124 @@ def index(request):
...
+            restart,
+            xmin,
+        )
+        if this_node:
+            if this_node['node_index'] == HEAD_NODE_INDEX:
+                # Check remote indexing
+                uuids_count = len(invalidated)
+                do_remote_indexing = _determine_indexing_protocol(request, uuids_count)
+                if did_timeout:
+                    log.warning(f"Remote indexing index: did_timeout.  Run on head node.")
```
hitz, line 568: so if the remote indexer goes down somehow it runs on itself?

src/snovault/elasticsearch/indexer.py
```diff
@@ -285,41 +458,124 @@ def index(request):
...
+                        delay_seconds=60,
+                        save_event=False
+                    )
+                    if status_code == 200:
+                        this_node['waiting_on_remote'] = True
+                        this_node['last_run_time'] = str(float(time.time()))
+                        indexer_state.put_obj(this_node['node_index'], this_node)
+                    else:
+                        log.warning(f"Remote indexing index: Failed sqs queue.  Run on head node.")
+                        result, indexing_update_infos = _run_indexing(*run_indexing_vars)
```
hitz, line 595: not a fan of * here.  It's somewhat obfuscatory.   Probably I would just pas a dict around but maybe there is a good reason not to do so.

hitz (changes requested):
OK, I think I got most of it, but since I wasn't that familiar with Tim's code this was pretty hard to follow.  The only major style thing is the two refactored _private functions need to be class methods or something and moved out of the main view method.  

The other major thing is in general that how the remote indexing works (which nodes are which, what are the configurable parameters, how does the SQS function...)  isn't documented, and it's not trivial to piece together from the code

src/snovault/elasticsearch/indexer_state.py
```diff
@@ -38,6 +46,100 @@ def includeme(config):
...
+        head_node['node_index'] = HEAD_NODE_INDEX
+        head_node['instance_name'] = instance_name
+        head_node['instance_state'] = instance_state
+        indexer_state.put_obj(head_node['node_index'], head_node)
+        # Create indexing node
+        indexer_name = f"{instance_name}{indexer_tag}"
+        indexer_instance_name, indexer_instance_state = _get_this_instance_name(
+            instance_name=indexer_name
+        )
+        if not instance_name:
```
pedrohr: should this be `if not indexer_instance_name:`?

src/snovault/elasticsearch/indexer_state.py
```diff
@@ -38,6 +46,100 @@ def includeme(config):
...
+            instance_name=head_node['instance_name']
+        )
+        if instance_name and instance_state:
+            head_node['instance_state'] = instance_state
+            indexer_state.put_obj(head_node['node_index'], head_node)
+        # Indexer
+        instance_name, instance_state = _get_this_instance_name(
+            instance_name=indexing_node['instance_name']
+        )
+        if instance_name and instance_state:
```
pedrohr, line 140: `_get_this_instance_name` might fail, in that case, should `did_fail = True` as we will not update the indexer_state

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+                label += ' - and shutting down indexer'
+                instance_name = this_node['instance_name']
+                remote_indexing = False
+                status_code = _send_sqs_msg(
+                    f"{instance_name}-cluster",
+                    remote_indexing,
+                    'Turn OFF remote indexer',
+                    delay_seconds=60,
+                    save_event=False
+                )
```
pedrohr: do we need to add error handling for this call?

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+        key: {'DataType': 'String', 'StringValue': val}
+        for key, val in msg_attrs.items()
+    }
+    queue_arn = _get_sqs_arn('ec2_scheduler_input')
+    response = _SQS.send_message(
+        QueueUrl=queue_arn,
+        DelaySeconds=delay_seconds,
+        MessageAttributes=msg_attrs,
+        MessageBody=body,
+    )
```
pedrohr: might be worth adding a try/catch/retry (maybe 3 times) until successful - but just a "nice to have", as the calls have error handling in the code

src/snovault/elasticsearch/indexer.py
caseylitton: yes, we should catch a failed sqs call.   This is were we need to wait for the instance to shut down before possibly allowing a follow up indexing too.   This should blocking.

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+                reset=True
+            )
+        elif other_node['started_indexing']:
+            label = 'Head node with remote indexing running'
+        elif this_node['waiting_on_remote']:
+            label = 'Head node waiting on remote indexing to start'
+            if this_time_in_state > 60*10:
+                # indexer machine did not start, move on without it
+                # need to cancel it?
+                did_timeout = True
```
caseylitton, line 287: Need to handle canceling the indexer instance.  The indexer could be coming up or the message in flight or not yet sent to the lambda function, etc.  

- Might be able to store the receipt handle in the node state and delete if possible. 
- We have the indexer aws instance state so check that too and maybe reset the clock if pending.
- Or just should reseting the index node state might handle it all.

This should be blocking too.

src/snovault/elasticsearch/indexer_state.py
caseylitton: It was in the indexer until adding the reset.  indexer.py imports indexer_state.py so we have to import from here.  

But yea the oop is a bit wonky.

src/snovault/elasticsearch/indexer.py
caseylitton: right, 10k might be low.

src/snovault/elasticsearch/indexer.py
```diff
@@ -44,6 +49,11 @@
 log = logging.getLogger('snovault.elasticsearch.es_index_listener')
 MAX_CLAUSES_FOR_ES = 8192
 DEFAULT_QUEUE = 'Simple'
+_SQS = boto3.client('sqs', region_name='us-west-2')
+_SQS_INPUT_TEMPATE = { # messageAttributes
+    'cluster_name': None,
+    'remote_indexing': None,
+}
```
paul-sud: Would be nice to see a class here. Not strictly necessary but makes the IDE/developer experience better, and makes future extension easier.

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
     return (related_set, False)
 
 
+def _determine_indexing_protocol(request, uuid_count):
+    remote_indexing = asbool(request.registry.settings.get('remote_indexing', False))
+    if not remote_indexing:
+        return False
+    default_threshold = 10000
```
paul-sud: This should be extracted to a constant somewhere, probably at the top of this file is OK.

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+    try:
+        remote_indexing_threshold = int(
+            request.registry.settings.get(
+                'remote_indexing_threshold',
+                default_threshold,
+            )
+        )
+    except:
```
paul-sud: Should try to avoid bare `except`. Looks like here you are trying to check if the conversion fails with `ValueError`, unless there is some failure mode of `request.registry.settings.get` I am missing:
```suggestion
    except ValueError:
```

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+    error_str = None
+    msg_attrs = copy.deepcopy(_SQS_INPUT_TEMPATE)
+    msg_attrs['cluster_name'] = cluster_name
+    msg_attrs['remote_indexing'] = '0' if remote_indexing else '1'
+    msg_attrs['save_event'] = '0' if save_event else '1'
+    msg_attrs = {
+        key: {'DataType': 'String', 'StringValue': val}
+        for key, val in msg_attrs.items()
+    }
+    queue_arn = _get_sqs_arn('ec2_scheduler_input')
```
paul-sud: What happens if `queue_arn` is None? This looks like it could happen if the queue doesn't exist as determined by `_get_sqs_arn`

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+        remote_indexing,
+        body,
+        delay_seconds=0,
+        save_event=False,
+    ):
+    error_str = None
+    msg_attrs = copy.deepcopy(_SQS_INPUT_TEMPATE)
+    msg_attrs['cluster_name'] = cluster_name
+    msg_attrs['remote_indexing'] = '0' if remote_indexing else '1'
+    msg_attrs['save_event'] = '0' if save_event else '1'
```
paul-sud: Any reason not to use booleans for these?

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+    if is_indexing_node:
+        this_node = indexing_node
+        other_node = head_node
+    if did_fail:
+        did_timeout = True
+        return this_node, other_node, continue_on, did_timeout
+    this_time_in_state = -1
+    other_time_in_state = -1
+    this_time_in_state = time_now - float(this_node['last_run_time'])
+    other_time_in_state = time_now - float(other_node['last_run_time'])
```
paul-sud: Duplicate declaration is unnecessary:
```suggestion
    this_time_in_state = time_now - float(this_node['last_run_time'])
    other_time_in_state = time_now - float(other_node['last_run_time'])
```

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+            this_node, other_node, time_now, did_fail, is_indexing_node = setup_indexing_nodes(
+                request,
+                indexer_state,
+                reset=True
+            )
+        elif other_node['started_indexing']:
+            label = 'Head node with remote indexing running'
+        elif this_node['waiting_on_remote']:
+            label = 'Head node waiting on remote indexing to start'
+            if this_time_in_state > 60*10:
```
paul-sud: Probably want to extract this constant to the top of the file so it's easier to find/change.

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+                # indexer machine did not start, move on without it
+                # need to cancel it?
+                did_timeout = True
+        else:
+            # Object updated to started_indexing right before indexing starts
+            label = 'Head node says no indexing going on'
+            if this_time_in_state > 10*60:
```
paul-sud: Same as L248, should extract to a top-level variable: https://github.com/ENCODE-DCC/snovault/blob/e504b02e2ad5d931b0dd84f623683d077eb0a7cd/src/snovault/elasticsearch/indexer.py#L248

src/snovault/elasticsearch/indexer.py
```diff
@@ -143,140 +154,302 @@ def get_related_uuids(request, es, updated, renamed):
...
+                label += ' - and shutting down indexer'
+                instance_name = this_node['instance_name']
+                remote_indexing = False
+                status_code = _send_sqs_msg(
+                    f"{instance_name}-cluster",
+                    remote_indexing,
+                    'Turn OFF remote indexer',
+                    delay_seconds=60,
```
paul-sud: How did you get this number? I would like to see a comment in the code explaining the choice.

src/snovault/elasticsearch/indexer_state.py
```diff
@@ -38,6 +46,100 @@ def includeme(config):
     config.add_route('_indexer_state', '/_indexer_state')
     config.scan(__name__)
 
+
+def _get_this_instance_name(instance_name=None):
```
paul-sud, line 51: I'm wondering if this is really the best place for this function, it seems pretty generic. Just a thought, I'm not opposed to leaving it here since it's only used for indexing.

src/snovault/elasticsearch/indexer_state.py
```diff
@@ -38,6 +46,100 @@ def includeme(config):
...
+                    instance_name = tag_obj.get('Value')
+                    break
+    return instance_name, instance_state
+
+
+def setup_indexing_nodes(request, indexer_state, reset=False):
+    time_now = float(time.time())
+    did_fail = False
+    is_indexing_node = False
+    node_state_template = {
```
paul-sud, line 80: It would be good to put this into a class.

paul-sud (changes requested):

src/snovault/elasticsearch/indexer.py
caseylitton: This was copied from the lambda function code where it was used in multiple files.  I remove it here since we only used it one.

src/snovault/elasticsearch/indexer.py
caseylitton: haha, yea.  The message attributes for sqs can only be certain data types.  Boolean is not an option. 

They're listed here if you search for MessageAttributes: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html

I just stuck with strings, but there are other options include a custom type I didn't look into.

src/snovault/elasticsearch/indexer.py
caseylitton: yes

src/snovault/elasticsearch/indexer.py
caseylitton: yes, there is some duplication here and other places when dealing with the states.  Fiddling with it may cause bugs at this point.  It's pretty stable at the moment even with the deficiencies.  We discussed an overhaul of the indexer. Which is a better use of time than cleaning the thing I just tacked on here.

caseylitton (commented):
Updated a bunch, I didn't handle any major refactor or design comments.  Concentrated on catching cases, bugs, and small clean ups.

src/snovault/elasticsearch/indexer.py
```diff
@@ -49,11 +51,14 @@
...
-    'remote_indexing': None,
-}
+# Allow three minutes for indexer to stop before head indexer process continues
+_REMOTE_INDEXING_SHUTDOWN_SLEEP = 3*60  
+# Allow indexer to remain up for 30 minutes after indexing before being shutdown
+_REMOTE_INDEXING_SHUTDOWN = 30*60
+# Allow indexer 20 minutes to startup before head indexer thread takes over indexing
+_REMOTE_INDEXING_STARTUP = 20*60
+# The number of uuids needed to start a remote indexer
+_REMOTE_INDEXING_THRESHOLD = 10000
```
hitz, line 61: Thanks this is really helpful!

src/snovault/elasticsearch/indexer.py
```diff
@@ -158,58 +163,61 @@ def _determine_indexing_protocol(request, uuid_count):
...
-        DelaySeconds=delay_seconds,
-        MessageAttributes=msg_attrs,
-        MessageBody=body,
-    )
-    log.warning(f"Remote indexing _send_sqs_msg: {body}")
-    status_code = response['ResponseMetadata']['HTTPStatusCode']
-    return status_code
+    boto_session = boto3.Session(region_name=AWS_REGION)
+    sqs_resource = boto_session.resource('sqs')
+    watch_dog = 3
```
hitz, line 203: rawf!

src/snovault/elasticsearch/indexer.py
```diff
@@ -204,12 +207,14 @@ def _send_sqs_msg(
         try: 
             queue_arn = sqs_resource.get_queue_by_name(QueueName='ec2_scheduler_input')
             if queue_arn:
+                print(queue_arn.url)
```
hitz: is this a debug print?

src/snovault/elasticsearch/indexer.py
```diff
@@ -204,12 +207,14 @@ def _send_sqs_msg(
...
                 response = sqs_resource.meta.client.send_message(
                     QueueUrl=queue_arn.url,
                     DelaySeconds=delay_seconds,
                     MessageAttributes=msg_attrs,
                     MessageBody=body,
                 )
+                print(response)
```
hitz: is this  a debug print?

hitz (changes requested):
This looks good except for the stray print statements... those should be removed or go into a Logging.

src/snovault/elasticsearch/indexer.py
caseylitton: removed, looked for more but didn't find any

hitz (approved):

paul-sud (approved):

PR #285, opened 2020-07-31
SNO-163-install-graphviz

caseylitton:
Quick fix for adding the svg back in and checking if dot was installed.

See encd ticket:
https://github.com/ENCODE-DCC/encoded/pull/3460

### DIFF
--- a/src/snovault/schema_graph.py
+++ b/src/snovault/schema_graph.py
@@ -1,15 +1,23 @@
+import logging
+
+from subprocess import (
+    Popen,
+    PIPE,
+    call as subprocess_call
+)
+
 from collections import defaultdict
 from past.builtins import basestring
 from pyramid.response import Response
 from pyramid.view import view_config
-from subprocess import Popen, PIPE
 from xml.sax.saxutils import quoteattr, escape
 from snovault import TYPES
 
+log = logging.getLogger(__name__)
 
 def includeme(config):
-    config.add_route('graph_dot', '/profiles/graph.dot')
     config.add_route('graph_svg', '/profiles/graph.svg')
+    config.add_route('graph_dot', '/profiles/graph.dot')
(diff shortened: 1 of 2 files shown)


keenangraham:
```
src/snowflakes/tests/test_graph.py::test_graph_svg FAILED                [ 36%]
________________________________ test_graph_svg ________________________________

testapp = <webtest.app.TestApp object at 0x7f501e5cc850>

    def test_graph_svg(testapp):
        res = testapp.get('/profiles/graph.svg', status=200)
>       res_json = res.json

src/snowflakes/tests/test_graph.py:9:
```
Seems like a real failure. Should same tests to encoded suite so they run from there (there are a number of duplicate tests already etc.).

src/snovault/schema_graph.py
```diff
@@ -84,8 +92,17 @@ def schema_dot(request):
 
 @view_config(route_name='graph_svg', request_method='GET')
 def schema_svg(request):
-    dot = digraph(request.registry[TYPES].by_item_type, request.params.getall('exclude'))
-    p = Popen(['dot', '-Tsvg'], stdin=PIPE, stdout=PIPE, stderr=PIPE)
-    svg, err = p.communicate(dot.encode('utf-8'))
-    assert p.returncode == 0, err.decode('utf-8')
-    return Response(svg, content_type='image/svg+xml', charset='utf-8')
+    try:
+        return_code = subprocess_call(['which', 'dot'])
```
keenangraham, line 96: Having the application check its dependencies seems a bit odd in this context. We should control the dependencies through install scripts and check functionality with view tests.

caseylitton: well, the view is calling the command.  Shouldn't it handle the scenario when the command doesn't exist?

keenangraham (approved):

PR #294, opened 2020-10-02
SNO 3027 kibana monitoring post refactor

Bek:


src/snovault/elasticsearch/searches/queries.py
```diff
@@ -94,7 +94,7 @@ def _get_or_create_search(self):
             self.search = Search(
                 using=self._get_client(),
                 index=self._get_index(),
-            )
+            ).extra(track_total_hits=True).query()
```
keenangraham, line 97: Think the ES7 upgrade stuff needs to be split into its own branch, but will make a note about this now. 

This should be set in an independent step to keep the coupling between the base query object and the steps that build up that query object loose (we shouldn't force the children queries to always report exact hits in other words). You can follow same pattern of adding an `.extra()` parameter as is done for `source` here: https://github.com/ENCODE-DCC/snovault/blob/3c1e3e3856ddea571eedbbecfe52218430cada94/src/snovault/elasticsearch/searches/queries.py#L831

Then add that step (called something like `self.add_total_hits()`) to the query building instructions here: https://github.com/ENCODE-DCC/snovault/blob/3c1e3e3856ddea571eedbbecfe52218430cada94/src/snovault/elasticsearch/searches/queries.py#L871

Also I don't think we want to call `.query()` here unless the DSL has somehow changed. You can see it is called explicitly in other places e.g.: https://github.com/ENCODE-DCC/snovault/blob/3c1e3e3856ddea571eedbbecfe52218430cada94/src/snovault/elasticsearch/searches/queries.py#L706

keenangraham (changes requested):

src/snovault/elasticsearch/searches/queries.py
Bek: I will get the changes requested in this branch: https://github.com/ENCODE-DCC/snovault/tree/SNO-172-ES7-upgrade with just the upgrade related commit. The DSL has changed in a way that breaks the test assertions. Without calling the `.query()`, top level field, `"query": {}` part of the overall query object isn't returned. Let me know which you prefer changed, test assertions or the parent class here.

keenangraham: ```Empty Search no longer defaults to match_all query and instead leaves the query key empty. This is backwards incompatible when using suggest.```

Kk. My guess is that we want to update the tests that expect this default to be returned, while making sure the tests that explicitly add a query still do so.

PR #295, opened 2020-10-05
SNO-124-make-elasticsearch-shards-configurable 

Bek:


keenangraham (approved):

base.ini
```diff
@@ -5,6 +5,8 @@ retry.attempts = 3
 file_upload_bucket = snowflakes-files-dev
 file_upload_profile_name = snowflakes-files-upload
 elasticsearch.server = localhost:9201
+elasticsearch.shards.primary = 1
+elasticsearch.shards.replicate = 0
```
hitz, line 9: @keenangraham @Bek I think we should start with adding the ability to configure the shards without actually changing it.

hitz (changes requested):

base.ini
Bek: With version 5, we don’t have the total shards limit and we can leave the default number as before. We should still change it to minimum for single machine demos, what do you think? I will look to it.

hitz: I don't think it matters much on a single-node demo; they already index just as fast and they have a low search load.  What difference do you think it would make?

PR #296, opened 2020-10-05
SNO-172-upgrade-to-elasticsearch7

Bek:


src/snovault/elasticsearch/create_mapping.py
```diff
@@ -133,10 +134,11 @@ def schema_mapping(name, schema):
             'type': field_type
         }
 
-        # these fields are unintentially partially matching some small search
-        # keywords because fields are analyzed by nGram analyzer
-        if name in NON_SUBSTRING_FIELDS:
-            sub_mapping['include_in_all'] = False
+        if name not in NON_SUBSTRING_FIELDS + TEXT_FIELDS:
```
keenangraham: I think we do want to copy the `TEXT_FIELDS` to `_all`. The search code just looks in `_all` field now and expects it to be analyzed. We definitely want description etc. searchable.

src/snovault/elasticsearch/create_mapping.py
```diff
@@ -167,12 +169,13 @@ def index_settings():
         'settings': {
             'index.max_result_window': 99999,
             'index.mapping.total_fields.limit': 5000,
-            'index.number_of_shards': 5,
-            'index.number_of_replicas': 2,
+            'index.number_of_shards': 1,
+            'index.number_of_replicas': 0,
```
keenangraham, line 154: Why are we lowering these?

src/snovault/elasticsearch/create_mapping.py
```diff
@@ -273,17 +276,14 @@ def audit_mapping():
...
-            'analyzer': 'snovault_search_analyzer'
-        },
         'dynamic_templates': [
             {
                 'template_principals_allowed': {
                     'path_match': "principals_allowed.*",
                     'match_mapping_type': "string",
                     'mapping': {
                         'type': 'keyword',
+                        'copy_to': '_all',
```
keenangraham: It doesn't make sense to me to copy `principals_allowed.*` to `_all`. I'm actually kind of surprised that the `links` and `unique_keys` fields are stored in `_all` in our current version, if that's what's happening.

src/snovault/elasticsearch/indexer.py
```diff
@@ -135,13 +135,11 @@ def get_related_uuids(request, es, updated, renamed):
                     {
                         'terms': {
                             'embedded_uuids': updated,
-                            '_cache': False,
```
keenangraham, line 138: Is there another option here? We probably don't want to worry about cache invalidation when we're finding related uuids for indexing.

src/snovault/elasticsearch/indexer.py
```diff
@@ -135,13 +135,11 @@ def get_related_uuids(request, es, updated, renamed):
...
                         },
                     },
                     {
                         'terms': {
                             'linked_uuids': renamed,
-                            '_cache': False,
```
keenangraham, line 144: Same as above.

src/snovault/elasticsearch/searches/queries.py
```diff
@@ -850,6 +850,9 @@ def add_sort(self):
                 *sort_by
             )
 
+    def add_exact_counting(self):
```
keenangraham, line 858: Nice! You'll have to add an `add_exact_counting` test to continue our 100% coverage of this code. Can see https://github.com/ENCODE-DCC/snovault/blob/79b5524ce9b9d7f7d32a1dbc143230f7efc5873a/src/snovault/tests/test_searches_queries.py#L2905 etc. for pattern.

keenangraham (changes requested):

src/snovault/elasticsearch/create_mapping.py
Bek: This is to make the search results as comparable as possible. Without the change, and without biosample term name filtering, the Elasticsearch 7 is going to show this experiment for "brain", "blood", or "breast" keywords  due to both words happening in the award description: https://www.encodeproject.org/experiments/ENCSR003IGU/?format=json. The result counts will be off.

src/snovault/elasticsearch/create_mapping.py
Bek: The old setting resulted in more than 1000 shards and hit the limit set by the Elasticsearch. The limit can be overridden but is not recommended. Regardless, this is changed in favor of [configurable shards](https://github.com/ENCODE-DCC/snovault/pull/295/files#diff-13be881525721a1251be911ff12f4c7a722ced10eb689a96d8c5821e8d9cd0e8L165) which also applies the same setting for single demos.

src/snovault/elasticsearch/create_mapping.py
keenangraham: `The result counts will be off.` Are you sure? I know description is currently included in `_all` so it's free text searchable. 

For example putting `RNA Evaluation human tissue W72 left colon microRNA-seq from Mortazavi` into search box returns the experiment with that description: https://www.encodeproject.org/search/?searchTerm=RNA+Evaluation+human+tissue+W72+left+colon+microRNA-seq+from+Mortazavi

How will we maintain this functionality with ES7?

src/snovault/elasticsearch/create_mapping.py
keenangraham: Could you clarify the new settings for local, single node demos, and production clusters?

src/snovault/elasticsearch/indexer.py
Bek: They have deprecated `_cache` field since 2.0 and have been raising error only after 5.x. It would be hard to cause cache invalidation on-demand during indexing and observe the effect. [More here](https://www.elastic.co/blog/better-query-execution-coming-elasticsearch-2-0)

src/snovault/elasticsearch/indexer.py
Bek: https://github.com/ENCODE-DCC/snovault/pull/296#discussion_r503997315

src/snovault/elasticsearch/indexer.py
keenangraham: Cool, so this shouldn't change behavior.

src/snovault/elasticsearch/create_mapping.py
Bek: You are right, the long query doesn't return if `TEXT_FIELD` is removed from `_all`. I will bring it back, build demo, and we can compare queries.

src/snovault/elasticsearch/create_mapping.py
Bek: Without the configurable shards branch, this new settings with 1 primary and 0 replicate will be applied in all cluster types. Old settings won't work in any cluster. With the configurable shards feature, local and single node demo clusters will have 1 primary and 0 replicate. Production cluster will have 3 primary and 1 replicate by default and is configurable at the deploy script.

src/snovault/elasticsearch/searches/queries.py
Bek: https://github.com/ENCODE-DCC/snovault/blob/2b86c6e7820979de77989c54a12d57fa03298cb2/src/snovault/tests/test_searches_queries.py#L2903

src/snovault/tests/test_searches_queries.py
```diff
@@ -2928,6 +2900,16 @@ def test_searches_queries_abstract_query_factory_add_source_object(dummy_request
     assert all([e in actual for e in expected])
     assert len(expected) == len(actual)
 
+def test_searches_queries_abstract_query_factory_add_exact_counting(params_parser, mocker):
+    from snovault.elasticsearch.searches.queries import AbstractQueryFactory
+    aq = AbstractQueryFactory(params_parser)
+    mocker.patch.object(AbstractQueryFactory, '_get_index')
+    AbstractQueryFactory._get_index.return_value = 'snovault-resources'
+    aq = AbstractQueryFactory(params_parser)
+    aq.add_source()
```
keenangraham: Can get rid of this part, only needed for `add_source()`.

src/snovault/elasticsearch/create_mapping.py
Bek: here is a demo for comparison: https://elasticsearch7.demo.encodedcc.org/

src/snovault/tests/test_searches_queries.py
Bek: sure thing, we will loose the first assertion, which is probably less important than it looks. This line is required I believe: `mocker.patch.object(AbstractQueryFactory, '_get_index')`

src/snovault/elasticsearch/create_mapping.py
Bek: This demo when indexed should be a closer match with the old one: https://elasticsearch7-similar-results.demo.encodedcc.org/

base.ini
```diff
@@ -53,13 +53,13 @@ set queue_worker_processes = 16
...
 set queue_worker_batch_size = 2000000
 
-[composite:regionindexer]
-use = egg:snovault#indexer
-app = app
-path = /index_file
-timeout = 60
-set embed_cache.capacity = 5000
-set regionindexer = true
+#[composite:regionindexer]
```
keenangraham: I think really delete it if we're getting rid of it.

development.ini
```diff
@@ -37,8 +37,8 @@ use = egg:rutter#urlmap
 [composite:indexer]
 use = config:base.ini#indexer
 
-[composite:regionindexer]
-use = config:base.ini#regionindexer
+#[composite:regionindexer]
```
keenangraham: Delete.

src/snovault/elasticsearch/indexer.py
```diff
@@ -387,7 +385,6 @@ def _load_indexing(request, session, connection, indexer_state):
             else:
                 status = request.registry[ELASTIC_SEARCH].get(
                     index=request.registry.settings['snovault.elasticsearch.index'],
-                    doc_type='meta',
```
keenangraham, line 390: Where does this get stored now? The `doc_type` concept is deprecated?

src/snovault/elasticsearch/indexer.py
```diff
@@ -535,7 +530,7 @@ def _run_indexing(
         request.registry[ELASTIC_SEARCH].indices.refresh(RESOURCES_INDEX)
         if flush:
             try:
-                request.registry[ELASTIC_SEARCH].indices.flush_synced(index=RESOURCES_INDEX)  # Faster recovery on ES restart
+                request.registry[ELASTIC_SEARCH].indices.flush(index=RESOURCES_INDEX)  # Faster recovery on ES restart
```
keenangraham, line 533: Can you explain difference?

src/snovault/elasticsearch/create_mapping.py
```diff
@@ -88,11 +84,16 @@ def schema_mapping(name, schema):
             mapping = schema_mapping(k, v)
             if mapping is not None:
                 properties[k] = mapping
-        return {
-            'type': 'object',
-            'include_in_all': False,
-            'properties': properties,
```
keenangraham, line 94: If this always had `properties` why the if/else now?

src/snovault/elasticsearch/create_mapping.py
```diff
@@ -133,10 +134,6 @@ def schema_mapping(name, schema):
             'type': field_type
         }
 
-        # these fields are unintentially partially matching some small search
-        # keywords because fields are analyzed by nGram analyzer
-        if name in NON_SUBSTRING_FIELDS:
```
keenangraham, line 138: No equivalent replacement of `copy_to: _all` here?

keenangraham (changes requested):
This looks close, just a couple of questions.

Also looks like real test failure where `cherry^` is no longer returning results (404). I noticed the same query between demo and prod also returning less `cherry^` results (looks like Files specifically)..

src/snovault/elasticsearch/indexer.py
Bek: `flush_synced` is deprecated

src/snovault/elasticsearch/create_mapping.py
Bek: if there is no mapping directive exists in the schema, object should be stored without its fields being analyzed. `else` clause prevents unnecessary dynamic mapping and `update_mapping` events in elasticsearch.

src/snovault/elasticsearch/indexer.py
Bek: yes, deprecated. Index is the doc_type for us now.

src/snovault/elasticsearch/create_mapping.py
Bek: Here is the newer commit: https://github.com/ENCODE-DCC/snovault/commit/afbe4761b19a10de2b1f27e505e971b3a3740ac9#diff-13be881525721a1251be911ff12f4c7a722ced10eb689a96d8c5821e8d9cd0e8R140

Bek (commented):
> This looks close, just a couple of questions.
> 
> Also looks like real test failure where `cherry^` is no longer returning results (404). I noticed the same query between demo and prod also returning less `cherry^` results (looks like Files specifically)..

we are back to the same assertions in the test and `cherry^` is returning 1 result as before in the tests. With the latest change, the File count is equal now.

base.ini
Bek: ok, will delete in encoded as well

src/snovault/elasticsearch/create_mapping.py
```diff
@@ -80,19 +73,27 @@ def schema_mapping(name, schema):
...
+        return schema_mapping(name,
+                              schema['items'],
+                              depth=depth+1,
+                              parent=type_)
 
     if type_ == 'object':
         properties = {}
         for k, v in schema.get('properties', {}).items():
-            mapping = schema_mapping(k, v)
+            mapping = schema_mapping(k, v, depth=depth+1)
```
keenangraham: `depth+1` -> `depth + 1`

src/snovault/elasticsearch/create_mapping.py
```diff
@@ -80,19 +73,27 @@ def schema_mapping(name, schema):
 
     # Elasticsearch handles multiple values for a field
     if type_ == 'array':
-        return schema_mapping(name, schema['items'])
+        return schema_mapping(name,
```
keenangraham:
```
return schema_mapping(
    name,
    schema['items'],
    depth=depth + 1,
    parent=type_
)
```

src/snovault/elasticsearch/create_mapping.py
```diff
@@ -133,10 +132,12 @@ def schema_mapping(name, schema):
             'type': field_type
         }
 
-        # these fields are unintentially partially matching some small search
-        # keywords because fields are analyzed by nGram analyzer
-        if name in NON_SUBSTRING_FIELDS:
-            sub_mapping['include_in_all'] = False
+        if name not in NON_SUBSTRING_FIELDS:
+            if depth == 1 or (depth == 2 and parent == 'array'):
```
keenangraham: I think this is the logic I'm most uncomfortable with, though the results certainly look closer than before. What's the implication of not tracking depth, as it didn't seem like it was tracked before? Is it possible to always copy to _all `if name not in NON_SUBSTRING_FIELDS`?

keenangraham (changes requested):

src/snovault/elasticsearch/create_mapping.py
Bek: Older demos showed that setting `"copy_to": "_all"` always here is equivalent to boosting every single field in the nested objects (minus the `NON_SUBSTRING_FIELDS`). Depth tracking stops `"copy_to": "_all"` after level 1. Before, if file or experiment award description has "brain" and "blood" in the text, the files or experiments show up under either "blood" or "brain" search term. Non-specific experiments that showed up in the results in earlier demos were due to "copy_to" being applied to the fields of deeply nested objects. Older `include_in_all` was applied by default and depth tracking is applied under the hypothesis that, this default application of the setting was weighed more towards top fields rather than nested object fields. Result is that, now all top fields of the object has `copy_to_all`, and nested fields need to be boosted like so `"file.award.description": 1.0`. We could have achieved the same matching results by boosting all of the individual fields of a type also without depth tracking, which is still an option but may take more effort to match results.

keenangraham: It seems like maybe defining in boost would make sense given we already have top-level fields there, e.g.:
 ```
"boost_values": {
        "accession": 20.0,
        "@type": 1.0,
        "alternate_accessions": 1.0,
        "assay_term_name": 20.0,
        "assay_term_id": 1.0,
        "assay_title": 20.0,
        "assay_slims": 5.0,
        "dbxrefs": 1.0,
        "aliases": 1.0,
        "biosample_ontology.term_id": 1.0,
        "biosample_ontology.term_name": 10.0,
```

> this default application of the setting was weighed more towards top fields rather than nested object fields.

I don't think this explains the case of award.description with blood or brain. Even if the weight was low on these they would still show up in results, though at the bottom. However we're not seeing them at all on current prod.

My understanding is that if `_all` is enabled on old ES then `include_in_all` is True by default. However you can set it to `include_in_all: False` for a certain field and it will apply to all subfields unless otherwise specified. I think looking at new and old mappings informative here. For file:

Screen Shot 2020-10-27 at 11 26 38 AM

Screen Shot 2020-10-27 at 11 27 55 AM

I'm a little concerned since the default has changed from don't include unless specifically set to always include if in first level (probably explains why we are seeing more hits on new demo?). The boost offers more specific control, while always including top-level fields doesn't.

In any case might be close enough for our current purposes, and could be tweaked later. But I think worth looking more into.

File: src/snovault/elasticsearch/create_mapping.py (hunk 12 above)
Comment (keenangraham) [reply src/snovault/elasticsearch/create_mapping.py:94]: This doesn't seem to be working properly for some fields, e.g. `cloud_metadata` in files. (Might have something to do with it being a calculated_property.)

Screen Shot 2020-10-27 at 12 18 01 PM

Screen Shot 2020-10-27 at 12 18 19 PM

Before and after. Now you can't look up cloud_metadata.url=* for example.

**Review by keenangraham — CHANGES_REQUESTED** [review]

**Comment by keenangraham** [comment]
One other thing. I suppose we should take a look at the dynamic_template definitions one more time. Looking at the original mapping definition I guess these are included in _all by default. I don't think being able to search principals_allowed is that useful but maybe links and unique_keys would be. 

```
'dynamic_templates': [
            {
                'template_principals_allowed': {
                    'path_match': "principals_allowed.*",
                    'match_mapping_type': "string",
                    'mapping': {
                        'type': 'keyword',
                    },
                },
            },
            {
                'template_unique_keys': {
                    'path_match': "unique_keys.*",
                    'match_mapping_type': "string",
                    'mapping': {
                        'type': 'keyword',
                    },
                },
            },
            {
                'template_links': {
                    'path_match': "links.*",
                    'match_mapping_type': "string",
                    'mapping': {
                        'type': 'keyword',
                    },
                },
....
```

It does seem like `group.admin` and `system.Everyone` give search results on prod, which seems a bit silly right?

Screen Shot 2020-10-27 at 2 00 25 PM

Possibly this also explains the file type search differences before you added the depth tracking?

keenangraham:
It looks like links and unique_keys are also by default included in all (though not linked_uuids). Could also explain result differences before adding all top-level fields.

Screen Shot 2020-10-27 at 2 10 47 PM

Screen Shot 2020-10-27 at 2 10 42 PM

src/snovault/elasticsearch/create_mapping.py
Bek: Right, previous assumption was that if the field was important, it would be in the schema. This can be fixed, implication is that any other object inserted that doesn't have a mapping directive in the schema will have all of its field values be analyzed as keyword.

src/snovault/elasticsearch/create_mapping.py
Bek: You are right that boost offers more specific control. I find that it's not feasible to determine what to boost for the number of types we have multiplied by the number of top fields for each type, make demo for each combination, and compare results with the old ES. `ctcf` search bringing up `File` type results was due to unintentional boost on the `aliases` field. Determining which other objects and fields were important to match the existing result count in the old ES maybe challenging.

Bek:
> One other thing. I suppose we should take a look at the dynamic_template definitions one more time. Looking at the original mapping definition I guess these are included in _all by default. I don't think being able to search principals_allowed is that useful but maybe links and unique_keys would be.
> 
> ```
> 'dynamic_templates': [
>             {
>                 'template_principals_allowed': {
>                     'path_match': "principals_allowed.*",
>                     'match_mapping_type': "string",
>                     'mapping': {
>                         'type': 'keyword',
>                     },
>                 },
>             },
>             {
>                 'template_unique_keys': {
>                     'path_match': "unique_keys.*",
>                     'match_mapping_type': "string",
>                     'mapping': {
>                         'type': 'keyword',
>                     },
>                 },
>             },
>             {
>                 'template_links': {
>                     'path_match': "links.*",
>                     'match_mapping_type': "string",
>                     'mapping': {
>                         'type': 'keyword',
>                     },
>                 },
> ....
> ```
> 
> It does seem like `group.admin` and `system.Everyone` give search results on prod, which seems a bit silly right?
> 
> Screen Shot 2020-10-27 at 2 00 25 PM
> 
> Possibly this also explains the file type search differences before you added the depth tracking?

Right, some an existing test rely on searching user uuid, but we don't boost user uuid, so it was getting a hit based on `"principals_allowed.edit": ["group.admin", "userid.81a6cc12-2847-4e2e-8f2c-f566699eb29e"]` because it contained user uuid...

Bek:
> One other thing. I suppose we should take a look at the dynamic_template definitions one more time. Looking at the original mapping definition I guess these are included in _all by default. I don't think being able to search principals_allowed is that useful but maybe links and unique_keys would be.
> 
> ```
> 'dynamic_templates': [
>             {
>                 'template_principals_allowed': {
>                     'path_match': "principals_allowed.*",
>                     'match_mapping_type': "string",
>                     'mapping': {
>                         'type': 'keyword',
>                     },
>                 },
>             },
>             {
>                 'template_unique_keys': {
>                     'path_match': "unique_keys.*",
>                     'match_mapping_type': "string",
>                     'mapping': {
>                         'type': 'keyword',
>                     },
>                 },
>             },
>             {
>                 'template_links': {
>                     'path_match': "links.*",
>                     'match_mapping_type': "string",
>                     'mapping': {
>                         'type': 'keyword',
>                     },
>                 },
> ....
> ```
> 
> It does seem like `group.admin` and `system.Everyone` give search results on prod, which seems a bit silly right?
> 
> Screen Shot 2020-10-27 at 2 00 25 PM
> 
> Possibly this also explains the file type search differences before you added the depth tracking?

File type searches not showing up was due to `aliases` field not being copied to `_all`. This brings up a worrisome implication for search, which is that if the file name didn't have `ctcf` in the string of it's name, then it wouldn't be included in the results even if the file was part of the experiment that targets `CTCF`.

src/snovault/elasticsearch/create_mapping.py
keenangraham: I think it's likely the difference is just due to not copying the `links` and `unique_keys` stuff to `_all`, which is by default `include_in_all: True` on old ES:
Screen Shot 2020-10-27 at 4 02 44 PM

Pretty sure it was the alias in unique keys that was matching `ctcf`, not the the alias in embedded properties. (Can confirm that alias in embedded properties is False for old mapping.)

It doesn't look like any of the embedded properties in old mapping are `include_in_all: True` unless they are defined in boost values for that schema. Could we go back to only copying embedded properties to _all if they are defined in boost, and just make the dynamic template links/unique keys have copy_to?

> ctcf search bringing up File type results was due to unintentional boost on the aliases field

Yeah, looks like only field ctcf is mentioned in file is alias and submitted_file name, but it's probably fine to have these intentionally searchable.

keenangraham:
> This brings up a worrisome implication for search, which is that if the file name didn't have `ctcf` in the string of it's name, then it wouldn't be included in the results even if the file was part of the experiment that targets `CTCF`.

Think this is fine. File for a long time didn't have any access to experiment biosample or assay information. Now that it does we should probably explicitly add these to boost values.

(In other words it's not that we expect `ctcf` to be a great search query for files, just that the conspicuous absence of files from new results was worth investigating.)

src/snovault/elasticsearch/create_mapping.py
Bek: Here is another demo with the changes: https://elasticsearch7-without-depth-tracking.demo.encodedcc.org/

PR #297, opened 2020-10-08
SNO-173 report clear indexer state

caseylitton:


src/snovault/elasticsearch/__init__.py
```diff
@@ -16,12 +18,79 @@
...
+            self.client = StrictRedis(
+                charset="utf-8",
+                decode_responses=True,
+                db=redis_db,
+                host='localhost',
+                port=6379,
+                socket_timeout=5,
+            )
+        else:
+            # ephemeral storage objects.
```
caseylitton, line 45: These objects only exist in the index endpoint.  We can't save the non persistent storage in the registry since the index endpoint is a different application as the _indexer_info endpoint.

development.ini
```diff
@@ -36,6 +36,7 @@ use = egg:rutter#urlmap
 
 [composite:indexer]
 use = config:base.ini#indexer
+set local_store = redis
```
caseylitton, line 39: If this variable is left empty then redis store is bypassed, but the stored state only exists for the life time of the index endpoint request.

src/snovault/elasticsearch/indexer.py
```diff
@@ -334,8 +338,99 @@ def _get_nodes(request, indexer_state):
...
+        'now_utc': str(datetime.datetime.utcnow()),
+        'now_pst': str(datetime.datetime.now(pytz.timezone('US/Pacific'))),
+        'indexer_info': 'not initialized',
+    }
+    result[f"events_{start}_to_{stop}"] = 'not initialized',
+    indexer_info = IndexerInfo(local_store='redis')
+    try:
+        result['indexer_info'] = indexer_info.hash_get(INDEX_ENDPOINT_INFO)
+        result[f"events_{start}_to_{stop}"] = indexer_info.get_indexing_events(start, stop)
+    except:
```
caseylitton, line 415: Could narrow this down, it does return a ConnectionError, but I'm not sure whether we want to keep the code that allows local storage to be turned off or require redis to be running.

src/snovault/elasticsearch/__init__.py
```diff
@@ -16,12 +18,79 @@
...
 
 ROOT_LOG = logging.getLogger('root')
 
 
+class LocalStoreClient():
+    client = None
+    
+    def __init__(self, redis_db=0, local_store=None):
+        if local_store:
+            self.client = StrictRedis(
```
pedrohr, line 36: we should have a fallback (try/catch) and default to the ephemeral storage objects if Redis fail. It will also prevent the class to throw exceptions up

src/snovault/elasticsearch/__init__.py
```diff
@@ -16,12 +18,79 @@
...
+from redis import StrictRedis
+
+
 PY2 = sys.version_info.major == 2
 
 
 ROOT_LOG = logging.getLogger('root')
 
 
+class LocalStoreClient():
```
pedrohr, line 31: I think this class should live in its own separate .py file outside of `__init__.py`

src/snovault/elasticsearch/__init__.py
```diff
@@ -16,12 +18,79 @@
...
+    client = None
+    
+    def __init__(self, redis_db=0, local_store=None):
+        if local_store:
+            self.client = StrictRedis(
+                charset="utf-8",
+                decode_responses=True,
+                db=redis_db,
+                host='localhost',
+                port=6379,
```
pedrohr, line 41: I know we don't have the same registry, is there any other config file we can read this from? if not, we should make this a constant

src/snovault/elasticsearch/indexer.py
```diff
@@ -334,8 +338,99 @@ def _get_nodes(request, indexer_state):
     return this_node, other_node, continue_on, did_timeout
 
 
+class IndexerInfo(LocalStoreClient):
```
pedrohr, line 341: I think we can move this class to a separate file, maybe the same file with `LocalStoreClient`

src/snovault/elasticsearch/indexer.py
```diff
@@ -334,8 +338,99 @@ def _get_nodes(request, indexer_state):
...
+                f"{event_tag}: Indexed {invalidated} uuids in {duration} seconds. Ended at {end}"
+            )
+        return events
+
+
+@view_config(route_name='_indexer_info', request_method='GET')
+def indexer_info(request):
+    events_range = request.params.get("events", '0:10')
+    try:
+        colon = events_range.index(":")
```
pedrohr, line 399: this is an internal endpoint so this suggestion might not be so important, but this throws an exception if the query string `events` has no ':' 
We should either return a BadRequest or default to '0:10' maybe

src/snovault/elasticsearch/indexer.py
pedrohr: if we fallback to the ephemeral storage in the `LocalStoreClient` we can remove this

caseylitton:
Moved to https://github.com/ENCODE-DCC/snovault/pull/299

PR #299, opened 2020-10-09
SNO-174-add-local-storage-redis

caseylitton:
SNO-175 is on a different PR but included for testing reasons

### DIFF
--- a/src/snovault/local_storage.py
+++ b/src/snovault/local_storage.py
@@ -1,0 +1,67 @@
+import binascii
+
+from os import urandom
+from datetime import datetime
+from pytz import timezone
+
+from redis import StrictRedis
+
+
+def base_result(local_store):
+    local_dt = datetime.now(timezone(local_store.local_tz))
+    return {
+        '@type': ['result'],
+        'utc_now': str(datetime.utcnow()),
+        'lcl_now': f"{local_store.local_tz}: {local_dt}",
+    }
+
+
+class LocalStoreClient():
+    '''
+    Light redis wrapper and redis examples
+    - get_tag function was added to return hex str
+    - Can access client directly for full functionality
+    '''
+    def __init__(self, **kwargs):
+        self.local_tz = kwargs.get('local_tz', 'GMT')
+        self.client = StrictRedis(
+            charset='utf-8',
+            decode_responses=True,
+            db=kwargs['db_index'],
+            host=kwargs['host'],
+            port=kwargs['port'],
+            socket_timeout=kwargs['socket_timeout'],
+        )
+
+    @staticmethod
+    def get_tag(tag, num_bytes=2):
+        '''
+        Tags are the tag plus a random hex bytes string
+        - Bytes string length is 2 * num bytes
+        '''
+        rand_hex_str = binascii.b2a_hex(urandom(num_bytes)).decode('utf-8')
+        return f"{tag}:{rand_hex_str}"
+
+    def ping(self):
+        return self.client.ping()
+
+    def dict_get(self, key):
+        return self.client.hgetall(key)
+
+    def dict_set(self, key, hash_dict):
+        return self.client.hmset(key, hash_dict)
+    
+    def get_tag_keys(self, tag):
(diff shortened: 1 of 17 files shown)


src/snovault/local_storage.py
```diff
@@ -0,0 +1,133 @@
...
+    LOCAL_STORAGE_EVENT_TAGS,
+    LOCAL_STORAGE_STATE,
+)
+
+
+def includeme(config):
+    config.scan(__name__)
+    config.add_route('local_storage', '/local_storage')
+    config.registry[LOCAL_STORAGE] = LocalStoreClient(
+        db=config.registry.settings['local_storage_db'],
```
pedrohr: this is correctly set as a registry setting, however, I wonder if there's any hidden pyramid magic that can give you the application name (main, indexer, regionindexer, etc...) and then have a hash map making the translation to the DB index number. If there isn't, maybe we could rename the variable to `local_storage_db_redis_index` or something like that, people might get lost when they see that for the first time. Or add a comment in the .ini files (maybe the comment is easier :))

src/snovault/local_storage.py
```diff
@@ -0,0 +1,133 @@
...
+        'utc_now': str(datetime.utcnow()),
+        'lcl_now': f"{local_tz}: {datetime.now(timezone(local_tz))}",
+        # Different pyramid process may have different registries, this returns memory address
+        'local_storage': str(local_store),
+    }
+
+    # Set/Get dict or hash
+    redis_info = local_store.dict_get(LOCAL_STORAGE_STATE)
+    if not redis_info:
+        redis_info = {
```
pedrohr: should this be part of the LocalStoreClient initializer, so we always have a state ready?

src/snovault/local_storage.py
```diff
@@ -0,0 +1,133 @@
...
+    # Set/Get single item
+    result['counter'] = int(local_store.item_get('counter')) + 1
+    local_store.item_set('counter', result['counter'])
+
+    # Store event key/value pairs with tag and list
+    event_key = request.params.get("event_key")
+    event_value = request.params.get("event_value")
+    event_tag = request.params.get("event_tag")
+    if event_key and event_value:
+        # Key values pairs are tagged for easier retrieval/organization
```
pedrohr: I think we should split this endpoint into two:
GET /events => get event info by tag or all events
POST /events => create a new event from a pair (event_key, event_value)
or any similar name for the endpoints :)

src/snovault/local_storage.py
```diff
@@ -0,0 +1,133 @@
...
+        if not event_tag:
+            # New tag created
+            event_tag = local_store.get_tag(4)
```
pedrohr: I would add a comment on the `4`. Bitwise operations can get tricky

src/snovault/local_storage.py
caseylitton: yea, the naming could be better.  There are a few others I might fix along with this one.

The only way I know to get the name of the section in the config is to add a var called name to the section in the config :)    We sorta do this in the ini files for encoded with the set indexer = true, set visindexer = true...at one point I named them explicitly.  Must have been in a feature that didn't get added.

src/snovault/local_storage.py
caseylitton: That's a good idea.  It's more static that way.  Also on initialization the connection will be checked.  Connection errors don't happen until a query is made.

src/snovault/local_storage.py
caseylitton: oh, and yea.  I think the db indexes should be global to all processes.  How would we get indexer data in the app if the app didn't know the indexer db index.

src/snovault/local_storage.py
caseylitton: these were moved SNO-173, this PR doesn't have an implementation of the local store.  It seemed like unnecessary code.

keenangraham:
Note to rebase this on dev before merge.

pedrohr (approved):
💯

PR #300, opened 2020-10-16
SNO-177-add-view-with-specified-fields

keenangraham:


src/snovault/auditor.py
```diff
@@ -244,8 +244,9 @@ def item_view_audit(context, request):
     path = request.resource_path(context)
     properties = request.embed(path, '@@object')
     inherit = context.audit_inherit
-    if context.embedded and '*' in context.audit_inherit:
-        inherit = context.embedded
+    embedded_paths = list(context.embedded) + [p.path for p in context.embedded_with_frame]
```
keenangraham: There are a couple places where we have to iterate over these paths. Think I could make this cleaner by just having a property that implements this logic once instead of leaking internals every time we have to do this.

pedrohr: does `context.embedded_with_frame` raise an exception if the context doesn't have it set?

pedrohr (approved):
💯

keenangraham:
Shouldn't have any functional change, put embedded and embedded_with_frame iteration logic in embedded_paths `@classmethod` (couldn't use `@property` because TypeInfo only interacts with class, not instance).

pedrohr (approved):

PR #301, opened 2020-10-21
SNO-175-fix-parametrize-unhashable-list-errors

caseylitton:
https://github.com/ENCODE-DCC/encoded/pull/3598


keenangraham (approved):
LGTM. Does pytest not use if you don't `auto-checkout`?

caseylitton:
This was due to a versioning issue.  this will become a problem when we update to pytest 6.1.1.

PR #308, opened 2020-11-11
Dockerfile for building snovault on ubuntu 18.04

annashch-insitro:
Dockerfile to build/run snovault on ubuntu 18.04


paul-sud:
Awesome work. If you're interested, I've installed `encodeD` (Snovault + ENCODE specific stuff) previously here: https://github.com/ENCODE-DCC/accession/blob/dev/docker/Dockerfile . There is some cruft in there specific to the use case for that image (it's not for production), but the general dependency installation is similar.

FWIW Elasticsearch 5.6 seems to work fine using the `apt` distribution, just need to add the official elasticsearch repos to the system.

PR #316, opened 2021-02-04
SNO-27-pip-installable

paul-sud:


keenangraham (approved):
Cool! Guess the `aws-ip-ranges` stuff is in encoded.

PR #321, opened 2021-03-02
SNO-192-update-pip-install

paul-sud:


setup.py
```diff
@@ -20,34 +18,25 @@
     "elasticsearch-dsl==5.4.0",
     "elasticsearch==5.4.0",
     "future==0.18.2",
-    "humanfriendly==6.1",
-    "jsonschema-serialize-fork @ git+https://github.com/lrowe/jsonschema_serialize_fork.git@2.1.1",  # noqa
+    "jsonschema_serialize_fork==2.1.1",
```
hitz, line 21: underscores not dashes?

hitz (approved):

setup.py
paul-sud: yes: https://pypi.org/project/jsonschema_serialize_fork/

PR #324, opened 2021-03-17
SNO-138-fix-pytest-warnings

paul-sud:


src/snovault/elasticsearch/searches/queries.py
```diff
@@ -198,13 +198,13 @@ def _get_collection_names_for_item_types(self, item_types):
...
-        return query.replace('/', '\/')
+        return query.replace('/', '\\/')
 
     def _escape_fuzzy_tilde(self, query):
-        return query.replace('~', '\~')
+        return query.replace('~', '\\~')
 
     def _escape_boost_caret(self, query):
-        return query.replace('^', '\^')
+        return query.replace('^', '\\^')
```
keenangraham, line 207: Which warnings does this fix? I'm a little worried we are actually changing the meaning of the query by adding additional backslash.

https://www.elastic.co/guide/en/elasticsearch/reference/5.6/query-dsl-query-string-query.html#_reserved_characters

https://lucene.apache.org/core/3_4_0/queryparsersyntax.html#Escaping%20Special%20Characters

These specific cases are to nullify characters in user input that could cause expensive queries/timeouts. By escaping the backslash itself doesn't it become the literal character interpreted by ES, while the `~` or `^` is again treated as a special character?

keenangraham (changes requested):

keenangraham:
I think maybe there's an issue with adding additional backslash to query string builder. The rest looks fine.

src/snovault/elasticsearch/searches/queries.py
paul-sud: You get an `invalid escape sequence` warning. Backslash is a reserved character in Python strings that should be escaped (unless you use raw string). The resulting strings with the change here are identical.

```
>>> "~".replace('~', '\~') == "~".replace('~', '\\~')
True
>>> "/".replace('/', '\/') == "/".replace('/', '\\/')
True
>>> "^".replace('^', '\^') == "^".replace('^', '\\^')
True
```

keenangraham: Ah, makes sense. Thanks for fixing.

keenangraham (approved):

PR #329, opened 2021-03-29
SNO-197-add-range-queries-and-min-max-aggs

keenangraham:


src/snovault/elasticsearch/searches/parsers.py
```diff
@@ -159,23 +166,45 @@ def get_must_not_match_filters(self, params=None):
...
+    def get_inequality_filters(self, params=None):
+        '''
+        Returns params with (lt:|lte:|gt:|gte:) syntax.
+        '''
+        return self.get_filters_by_condition(
+            key_and_value_condition=lambda _, v: bool(INEQUALITY_REGEX.match(v)),
+            params=params
+        )
+
+    def get_non_inequality_filters(self, params=None):
```
hitz, line 178: just a comment, this is a hilarious method name.   Should probably be dont_not_get_non_inequality_filters() :)

hitz (approved):

src/snovault/elasticsearch/searches/parsers.py
keenangraham: Yeah, I agree this is horrible. I actually wish there was a better name to distinguish `=` and `!=` operators from `>`, `>=`, `<`, `<=` operators (or even a name for these symbols besides angle brackets?). Relational or comparison operators seem to include all of them. Somehow `!=` is an equality operator and an inequality operator?

So `get_inequality_filters` as a substitute for _get anything with the `gt:, lt:, gte:, lte:` syntax_ definitely a compromise. And `get_non_inequality_filters` is sort of the consequence of matching the symmetry of `get_non_wildcard_filters` and avoiding the ambiguity of `get_equality_filters` which could imply getting filters with equal signs (including e.g. `file_size=lte:3000`, which is specifically what I want to filter out).

pedrohr (approved):
All good 👍 

I just have the same comments from https://github.com/ENCODE-DCC/encoded/pull/3812/files about following the tokenization -> constants pattern from Snovault. But I guess trying to refactor or change at this point in time might be too costly.

PR #336, opened 2021-04-16
SNO-201-add-search-config-registry

keenangraham:


src/snovault/elasticsearch/searches/configs.py
```diff
@@ -46,3 +55,188 @@ def __init__(self, allowed_kwargs=[], **kwargs):
...
+        return self.registry.get(name, default)
+
+    def _resolve_config_name(self, name, use_defaults=True):
+        if name in self.aliases:
+            yield from self._resolve_config_names(
+                self.aliases[name],
+                use_defaults=use_defaults
+            )
+        elif use_defaults and name in self.defaults:
+            yield from self._resolve_config_names(
```
pedrohr, line 150: if `name in self.defaults[name]` or `name in self.aliases[name]` we will get an infinite loop. It's something we can control in the definitions of the config dict, but should we add something preventing that to be set? this is not a blocker for now...

pedrohr (approved):
👍

PR #350, opened 2021-06-16
Update README.rst for java version consistency

cricketsloan:
When I was attempting to follow these instructions, I got an error that version 1.8 was not installed and Otto pointed out that it was version 11 that we just installed.  Changing the command to have -v 11 worked.  I am not sure where the 1.8 came from and wondered if possibly it was meant to be 11.8.  This fix has 11.


paul-sud (approved):
Looks fine. I just checked my setup and I don't even have `JAVA_HOME` set but the app works fine. But I think that's a different issue

PR #354, opened 2021-08-04
SNO-206-use-redis-cache-for-rnget-view

keenangraham:


src/snovault/elasticsearch/searches/decorators.py
```diff
@@ -96,3 +96,27 @@ def wrapper(*args, **kwargs):
...
+    using `key` function. Cache should implement mapping interface.
+    '''
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            if not condition(*args, **kwargs):
+                return func(*args, **kwargs)
+            cache_key = key(*args, **kwargs)
+            try:
+                return cache[cache_key]
```
pedrohr, line 112: I forgot the behavior of Redis in python, will it always raise an exception if the key is not set? I remember in Ruby it returns nil so this wouldn't work

pedrohr (approved):

PR #367, opened 2021-10-14
SNO-216 Fix dev-servers 100% cpu usage when nothing to print.

lrowe:
Add a 100ms sleep to avoid dev-servers spinning at 100% cpu when it's got nothing to print.


ottojolanki (approved):
LGTM

ottojolanki:
Confirmed that rebase did not change anything. Still good.

PR #370, opened 2021-10-21
SNO-217-register-search-config-pieces-from-item

keenangraham:


hitz (approved):

setup.py
```diff
@@ -35,7 +35,7 @@
     "redis==3.5.3",
     "requests==2.22.0",
     "simplejson==3.17.0",
-    "snovault-search==1.0.3",
+    "snovault-search @ git+https://github.com/ENCODE-DCC/snovault-search.git@register-search-config-pieces-from-item",
```
gabdank: @keenangraham after we merge this branch, and then want to release snovault 1.0.62 - I am unclear if this line should not change?

keenangraham: Yes, should release snovault-search first and point to it before merging this.

